PR #36, opened 2019-04-16
Fixes for loading texts and opt-out button bug

blaine-wataru:
- Fixed opt-out button bug, should function properly again.
- Moved loading texts from hard-coded strings to messages.properties file
- Removed loading texts from owners and preferences pages. I am aware Zak implemented this as well, his changes can overwrite what I did here when he merges, if he went into more detail with the functionality.


blaine-wataru:
Codacy error is false positive. Code was something else, needed to be changed to this for opt-out to work again.

PR #63, opened 2019-05-31
removed multi user include exclude buttons

zakgilbert:
Groupings include and exclude multiple user buttons have been removed


zakgilbert:
Just commented out html, shouldnt have any problems

PR #79, opened 2019-07-03
Change 'save' and 'cancel' icons

audreyford:


kahlinbaughman:
Was fixed and pushed from local machine.

PR #92, opened 2019-07-18
Changing more spacing, tightened and aligned table spacing

kahlinbaughman:


kahlinbaughman:
Was hoping to reproduce the problem.

PR #266, opened 2020-10-30
GROUPING - 681 

yiwenc22:
Added message to AddOwner, So it show a message when a empty string is added


zakgilbert:
Would you like me to merge this?

PR #334, opened 2021-05-05
Reimplemented problematic statement with hard-coded true to attempt fixing test

Kobeyarai808:


Kobeyarai808:
Determined with Michael that this hard-coded true is not going to resolve the issue and should be retracted.

PR #364, opened 2021-07-17
New getMemberships function returns different number of Memberships than Old function.-GROUPINGS-842

zakgilbert:
[GROUPINGS-842](https://www.hawaii.edu/jira/browse/GROUPINGS-842)


mhodgesatuh (approved):

mhodgesatuh:
Have you reviewed the checks failure report?

zakgilbert:
The error is with an ‘_’ That has always been there so I’m not sure why
it’s throwing the error now. I’ll look into it more.

PR #367, opened 2021-07-23
Dev-gavin4-851

devgav:
UI - Cleaned up syntax in HomeController.java, specifically changed @RequestMapping to @GetMapping and removed unused parameters from methods. Also made a @GetMapping change in GroupingRestController.java on line 214.


mhodgesatuh:
Some issues to review

- line 50, messages.properties: spacing around "=" is not consistent with the rest.  Might want to put it back the way it was.
- lines 1428, 1689, general.controller.js: best not to promoted commented-out code to the master branch.

fduckart:
Cool. There are some merge conflict markers still in there, though. You need to remove them. A bunch of the differences are just simple formatting changes, too, which appears to have complicated the merge. The style in there before was a bit off from our project style, which is why there were a lot more differences in the commit than the actually code changes you made. So... If you maintain the one style item where the previous committer aligned method parameters, you'd probably have less merge conflicts. Changing the style back to our project style is okay, too, though. 

Also, those things Michael noted.

PR #369, opened 2021-08-03
Implement a special message to display when you have 1 grouping left

devgav:
Implemented a way to display special message for the case where you are a owner and the case were you are an admin for when you have 1 grouping left. Also cleaned up some of the messages.properties file and the html code that had multiple span tags.


mhodgesatuh:
I notice in your latest commit you are using style=“…” code rather than classes.  This is not DRY and should be avoided.  Recommend you use classes such that you define the style once and use many times.  We can talk about this in the groupings channel or at a meeting.

PR #376, opened 2021-08-16
Implement a special message to display when you have 1 grouping left 

devgav:
Implemented a way to display special message for the case where you are a owner and the case were you are an admin for when you have 1 grouping left. I also cleaned up unused/extra span tag's in html files and in messages.properties


mhodgesatuh:
The use of style is not DRY.  Need to use classes.

        <span style="color: black">Are you sure you want to remove</span>
        <span style="font-weight: bold"> {{ userToRemove.name }}</span>
        <span style="color: black"> from the {{listName}} list?</span>

PR #381, opened 2021-09-09
GROUPINGS-879, 883, 890

nadine-alcantara:
Fixed missing groupings spinners, broken import file functionality, and presentation errors for modals


mhodgesatuh:
Thanks.  I reviewed and it looks good to go once you have presented to the team.  You might highlight some of the corrections for consistency to help the rest of the team get better at spotting 
  <th style="font-weight: normal"> (bad in our case)
    vs
  <th class="font-weight-normal"> (good in our case)

PR #383, opened 2021-09-10
Implemented remove button

mikel-ishihara:


fduckart:
I realize this is in your dev branch, but... the changes to importModal.html and general.controller.js do not appear to be AngularJS, so you should probably fix those parts.

PR #384, opened 2021-09-16
Create an endpoint that sends the number of memberships and groupings owned.

chakhon:
[GROUPINGS-840](https://www.hawaii.edu/jira/browse/GROUPINGS-840)


Kobeyarai808 (commented):
This looks great!

PR #388, opened 2021-09-18
Dev ishihara 884

mikel-ishihara:
This pull request if for tickets 884 and 892. For 884, I added a "Remove File" button that works by clearing the file input. I also updated the importModal.html to clarify what the remove button does, I would like second opinions on if it sounds fine. Ticket 892 is simply adding padding, as shown in the image attached to the ticket on the UH Kanban.

### DIFF
--- a/src/main/resources/static/css/main.css
+++ b/src/main/resources/static/css/main.css
@@ -373,6 +373,10 @@ table .p-12 {
     min-height: 25rem;
 }
 
+.column-padding {
+    padding-top: 20px;
+}
+
 .selected-grouping-card {
     border-bottom: 0;
     border-left: 0;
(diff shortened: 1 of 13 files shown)


fduckart:
Hey, can you do me a favor and squash these commits into a single one? Also, the code should use AngularJS, not what I see in there.

fduckart:
Also, this doesn't appear to be angularjs...

```
$scope.removeTextFile = function () {
    document.getElementById('upload').value = null;
 }
```

mikel-ishihara:
I have been trying to figure out how to squash my commits, but since I already pushed them, I don't believe I can squash these commits. I have fixed the problem of not using Angular in the most recent commit (I created a dev-ishihara-884-2 branch so that it would show as a single commit).

PR #389, opened 2021-09-21
Fix Failing Jasmine Tests-GROUPINGS-891

zakgilbert:
[GROUPINGS-891](https://www.hawaii.edu/jira/browse/GROUPINGS-891)

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -1613,25 +1613,6 @@
             $scope.currentPagePerson = 0;
         }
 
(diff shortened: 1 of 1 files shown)


fduckart:
Can you 'squash' the commits into a single commit? That objToPageArray function... think we need some tests for it. Also, you defined objToPageArray with two parameters, but then call it with only one argument.

fduckart:
Think it would be ideal if you could squash the commits.

zakgilbert:
You would rather have all changes on one commit?

fduckart:
Not necessarily. I guess group like things together.

PR #393, opened 2021-09-30
Groupings-907

zakgilbert:


fduckart:
Can you please fix a couple of a wart that was probably in there from the first template I created?  1) Try to get rid of the UserContextService#setCurrentUhUuid method.

Also, not sure how those uhUuid camel case changes got i there... I don't see them like that in master.

PR #395, opened 2021-10-05
Find another way to access $scope.currentUser

chakhon:


fduckart:
Hey, in GroupingsRestController, remove the parameter from your currentUsername method, autowire in UserContextService class into the class, then call it get the current user. And for currentUsername, you should be able to return the User class. You have a unusual hack in there now, I think because you tried to return the value in the ok builder method instead of the body method.

fduckart:
Also, forgot to ask... do you have the name of a html file that has that  'name' field on it so that I can look at it?

fduckart:
All these intermixed commits from various people just can't be right process-wise.

PR #403, opened 2021-10-13
getMemberAttributes return Person instead of Map-GROUPINGS-912

zakgilbert:
[GROUPINGS-912](https://www.hawaii.edu/jira/browse/GROUPINGS-912)


zakgilbert:
👍

PR #405, opened 2021-10-14
$scope.currentUser is being hydrated Incorrectly

chakhon:


mhodgesatuh:
badge token looks correct

PR #407, opened 2021-10-14
Remove inline styling and change CSS to SCSS

devgav:


mhodgesatuh:
I notice that codacy did not process your config file specifying an intent level of 4.  So I poked around in codacy and discovered that I can change it there.  And I did.  If we can figure out how to make codacy reevaluate we can determine it will honor the project's setting of 4.

fduckart:
Coming along nicely!

devgav:
> 
> 
> I notice that codacy did not process your config file specifying an intent level of 4. So I poked around in codacy and discovered that I can change it there. And I did. If we can figure out how to make codacy reevaluate we can determine it will honor the project's setting of 4.

devgav:
> 
> 
> I notice that codacy did not process your config file specifying an intent level of 4. So I poked around in codacy and discovered that I can change it there. And I did. If we can figure out how to make codacy reevaluate we can determine it will honor the project's setting of 4.

So it should be fine for me to delete the .stylintrc file? Also if it is possible is there a way to ignore that mainv2_3.css file since that shouldn't be looked at by Codacy, it just get's compiled by the mvn plugin.

mhodgesatuh:
I suppose so, unfortunately.  Codacy reports that it didn't find it.

PR #410, opened 2021-10-16
$scope.currentUser is being hydrated Incorrectly

chakhon:
Please double check that no changes has been reverted by me, thank you.


fduckart:
The file updates look good, but I'm thinking you should cancel this pull request and start over since it's such a bizarre mess. So, save away the updates somewhere so you don't lose them. Do a fresh pull. Then commit your changes on top of the clean pull. I'll expect a pull request with a single commit in it.

PR #416, opened 2021-10-20
Bump owasp-java-html-sanitizer version.

fduckart:


mhodgesatuh:
To clarify, the dev-2-1-1 branch is defunct, which is what PR is for, so all is well and the above is just a typo.

PR #426, opened 2021-10-29
Error message for adding invaild users to Owner

chakhon:
[GROUPINGS-920](https://www.hawaii.edu/jira/browse/GROUPINGS-920)

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -1065,6 +1065,7 @@
             if (_.isEmpty(ownerToAdd)) {
                 $scope.emptyInput = true;
             } else {
+                $scope.existsInGrouper(ownerToAdd, list);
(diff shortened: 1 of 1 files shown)


mhodgesatuh:
Please include Andre and Kobey as reviewers since they are both branch managers.

fduckart:
I think it would be awesome if you could take a crack at renaming the "existsInGrouper" function. It started life a year ago as a boolean-like function, but now as become a different beast altogether. A new name that didn't imply it was a boolean function would be the smallest, most helpful change, methinks,

chakhon:
> A new name that didn't imply it was a boolean function would be the smallest, most helpful change,

Created ticket [GROUPINGS-931](https://www.hawaii.edu/jira/browse/GROUPINGS-931) to address the issue.

ruizaj:
Looks good, you can go ahead with the merge.

PR #427, opened 2021-11-05
Dev ishihara 884

mikel-ishihara:


mikel-ishihara:
Added clear button and modified readTextFile function to use FileList array instead of nodes from the importModal.html

mhodgesatuh:
This sticks out:             let file = document.getElementById('upload').files[0];
If we are going to stick to angularJS then the above doesn't follow this principle.  Using a quick search I found this as a starting point: <https://stackoverflow.com/questions/18571001/file-upload-using-angularjs> for further reading.

And is there an angularjs alternative for this:
                angular.element(document.querySelector('#upload')).val(null);

PR #429, opened 2021-11-05
Removed some lgtm errors

lukepag:
Removed 
-Superfluous trailing arguments from general.controller.js
-Useless null check from CheckForPattern.java (This check is useless, dir cannot be null here, since new File(...) always is non-null.)


mhodgesatuh:
Does this fix need to be backported to 220?

ruizaj:
I'm merging this, looks pretty fine.

PR #432, opened 2021-11-19
908/921: Backported Changes 

ruizaj:


Kobeyarai808 (commented):
This looks good. @zakgilbert What do you think?

mhodgesatuh:
There's no harm in letting this go as is, but the description requires decoding.  A brief explanation of what's really going on is the ideal description and this should be in the additional info section: 908/921: Backported Changes #432

Here's an interesting article that goes a little too far for us, but it does go over the guiding principles for describing a commit.

zakgilbert (approved):

PR #433, opened 2021-11-19
Groupings-892

mikel-ishihara:
This is for adding padding above the reset grouping column in a grouping's preferences. This code is the same as the code that was pushed to dev-2-2-0 already for the same purpose.

### DIFF
--- a/src/main/resources/static/css/mainv2_3.css
+++ b/src/main/resources/static/css/mainv2_3.css
@@ -519,6 +519,10 @@ form.seamless input {
   }
 }
 
+.column-padding {
+    padding-top: 20px;
+}
+
 .modal-danger {
   color: #721c24;
 }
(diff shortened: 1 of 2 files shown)


mhodgesatuh:
likely this will also need to be ported to dev-2-2-0.

Kobeyarai808 (commented):
This looks good, no regressions.

fduckart:
You should use Bootstrap 4, eh.

mhodgesatuh:
Please create another ticket referencing this one and requesting that the next iteration of the code utilize bootstrap 4.  Thanks.

mhodgesatuh:
I plan on creating a new ticket to have a team member perform a comprehensive review of our code to see where we aren't, but should be using bootstrap.

fduckart:
You just need to put something like this in the class attribute: pt-4

See here: https://getbootstrap.com/docs/4.5/utilities/spacing/

PR #434, opened 2021-11-19
Groupings 884

mikel-ishihara:
This is the working DOM code for the clear button in the import modal. Michael said to push this working code and create a ticket for the future to change this code using AngularJS. There are no tests for this feature.

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -628,16 +628,14 @@
         };
 
         /**
-         * Read a text file(.txt) from client side. The file should consist of a list of UH user names or ids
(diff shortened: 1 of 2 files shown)


mhodgesatuh:
Does this also need to be applied to dev-2-2-0?

Kobeyarai808:
> Does this also need to be applied to dev-2-2-0?

My opinion to this is no, it should stick in 2-3-0 since it is a new feature. As of right now, the user can refresh the page to remove the file and upload another one. In my opinion, this is not fix to a critical bug that needs to go out in 2-2-0. 

Any thoughts? Other than that, I feel like this is good to commit to master branch. @ruizaj any thoughts?

mhodgesatuh:
Good point.  Future features are for the master branch only.  Thanks for catching that.

PR #437, opened 2021-11-29
Fix Remove button for removing users via textbox

zakgilbert:
Port changes from other branch.
[GROUPINGS-924](https://www.hawaii.edu/jira/browse/GROUPINGS-924).


mhodgesatuh:
One of the two Codacy issues might be easily fixed, but I'm not sure:

Screen Shot 2021-11-29 at 1 09 26 PM

zakgilbert:
Got rid of the _return_ and used an _if/else_ instead.

PR #438, opened 2021-11-30
Rename existsInGrouper to validateAndAddUser, fix bug where adding invalid user does not create error modal

chakhon:
[GROUPINGS-931](https://www.hawaii.edu/jira/browse/GROUPINGS-931)
[GROUPINGS-936](https://www.hawaii.edu/jira/browse/GROUPINGS-936)

Still need to port changes to dev-2-2-0


ruizaj:
Looks pretty good to me, no regressions so far.

PR #443, opened 2021-12-09
Fixed name attribute casing

nadine-alcantara:


ruizaj:
Looks great, I will be pushing this in.

PR #446, opened 2021-12-21
Add multi-remove feature to owners list #901

devgav:


devgav:
There is a codacy issue in the GroupingsRestControllerTest, it says that there needs to be a assert() or fail() in the test. We use a specific testing framework so I don't believe that codacy picks this up. 

Kind of a hack since the asserThat doesn't really test do anything but I remember getting this issue before and doing something like this: 
`MockMvc Result = MockMvc.perform(...)
assertThat(result, notNullValue());`

Kobeyarai808:
I am not too sure about how codacy checks this. Maybe @zakgilbert could help with this?

zakgilbert (approved):

fduckart:
This code looks pretty odd to my eyes...
   if (person.result === "SUCCESS" || person.resultCode === "SUCCESS")

fduckart:
Also... we've discussed the "JUnit tests should include assert" issue quite a number of times now... Michael keeps hinting that someone should add a comment so that we will remember what is going on here... but alas, it's not getting traction. I guess my comment regarding this codacy issue is to step back and think about what it is happening.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -1360,7 +1368,7 @@
         $scope.batchRemoveResponseHandler = function (response) {
             let success = false;
             for (let person of response) {
-                if (person.result === "SUCCESS") {
+                if (person.result === "SUCCESS" || person.resultCode === "SUCCESS") {
```
zakgilbert, line 1371: I message you about this on slack...

zakgilbert:
> Also... we've discussed the "JUnit tests should include assert" issue quite a number of times now... Michael keeps hinting that someone should add a comment so that we will remember what is going on here... but alas, it's not getting traction. I guess my comment regarding this codacy issue is to step back and think about what it is happening.

are you talking about this line?
`        assertThat(result, equalTo(result));` or just the fact that there is no usage of `assert` in a test?

devgav:
> This code looks pretty odd to my eyes... if (person.result === "SUCCESS" || person.resultCode === "SUCCESS")

This is because the object that is returned from removing a person from exclude/include is different than the removing an owner.
Here is the object the values that are returned from removing a person from include/exlcude:
image
and here is the object that is returned from removing a person from owners:
image

PR #455, opened 2021-12-30
Add multi-add feature to owners #696

devgav:
[GROUPINGS-696](https://www.hawaii.edu/jira/browse/GROUPINGS-696)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
@@ -401,8 +401,8 @@ public class GroupingsRestController {
      * read and write privileges
      * of a grouping.
      */
-    @PostMapping(value = "/{groupingPath}/{newOwner}/assignOwnership")
-    public ResponseEntity<String> assignOwnership(Principal principal,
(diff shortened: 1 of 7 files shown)


mhodgesatuh:
String uri = String.format(API_2_1_BASE + "/groupings/%s/multiAddOwners/%s", safeGrouping, safeNewOwner);
Seems to me we are losing sight of the RESTful API standard.  The fact that a put us being used already communicates that we are adding or updating.  So I suspect that /multiAddOwners should be /owners, but I haven't checked to see if there is a conflict such that we need this transitory endpoint until there's no difference between adding a single or multiple owners.

devgav:
> ```
>     String uri = String.format(API_2_1_BASE + "/groupings/%s/multiAddOwners/%s", safeGrouping, safeNewOwner);
> ```
> 
> Seems to me we are losing sight of the RESTful API standard. The fact that a put us being used already communicates that we are adding or updating. So I suspect that /multiAddOwners should be /owners, but I haven't checked to see if there is a conflict such that we need this transitory endpoint until there's no difference between adding a single or multiple owners.

the naming convention should be fixed both remove owners and add owners have the /owners now.

Kobeyarai808 (approved):
Looks good!

PR #457, opened 2022-01-04
2.2.0 ADA bugs /info page

chakhon:
[GROUPINGS-937](https://www.hawaii.edu/jira/browse/GROUPINGS-937)

### DIFF
--- a/src/main/resources/templates/info.html
+++ b/src/main/resources/templates/info.html
@@ -15,7 +15,7 @@
         <div class="container">
             <div class="row">
                 <div class="col-md-12">
-                    <h2 class="h1 text-center d-block">What is a UH Grouping?</h2>
+                    <h1 class="h1 text-center d-block">What is a UH Grouping?</h1>
                 </div>
                 <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


mhodgesatuh:
For my edification, what does changing role from "presentation" to "img" contribute to resolution of the ADA shortcoming?

mhodgesatuh:
And there's a plan to port this to master?

chakhon:
> For my edification, what does changing role from "presentation" to "img" contribute to resolution of the ADA shortcoming?

The [presentation role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) remove an elements implicit native semantics, I am not sure how that applies to an  element. The axe Dev tools tells us that "ARIA role presentation is not allowed for given element", so I replaced it with the "img" role instead. Perhaps I could use "aria-hidden" instead.

chakhon:
> And there's a plan to port this to master?

I will have to discuss this with Kobey and Zak.

zakgilbert (approved):

PR #462, opened 2022-01-05
2.2.0 ADA bugs /admin, Manage Person page

chakhon:
[GROUPINGS-947](https://www.hawaii.edu/jira/browse/GROUPINGS-947)

### DIFF
--- a/src/main/resources/templates/admin.html
+++ b/src/main/resources/templates/admin.html
@@ -154,7 +154,7 @@
                     </div>
                 </div>
                 <!-- Manage Person -->
-                <div class="tab-pane fade" id="manage-person" role="tab">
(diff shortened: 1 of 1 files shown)


Kobeyarai808 (approved):
Looks good!

zakgilbert (approved):

mhodgesatuh:
My understanding is the branch managers not only approve, but also handle the merge.  Is that something that we need to discuss further?

Kobeyarai808:
That was my original thought, it turned into the user merging because of the many hands touching each pull request. I oftentimes find myself reviewing the pull request before everyone has looked at it, thus leaving it for the others to look at. Did you think it's a better idea for us to merge it right when we approve it?

PR #488, opened 2022-01-25
Startup analyzer and handshake property option. String utilility clean up.

fduckart:


zakgilbert (approved):

zakgilbert:
Needs to be forward ported to 2.3 as well?

PR #491, opened 2022-01-25
Startup analyzer and handshake property option. String utilility clean up.

fduckart:


mhodgesatuh:
Typo: trunctate
"truncate"
Empty is also spelled wrong, but it is just a comment.  I would have seen it except that a trailing space was added to the line, which mean that the editor wasn't configured to remove trailing spaces.

Kobeyarai808:
Frank knows about the typo, he will go in and fix it with another pull request.

PR #495, opened 2022-01-25
fix console error warnings for dev-2-2-0

chakhon:
[GROUPINGS-986](https://www.hawaii.edu/jira/browse/GROUPINGS-986)

getNumberOfMemberships() uses loadData() to retrieve the membership number, but it takes a while to return the number from the API, so when we refresh the page before the number is returned, we get an error. I gave getNumberOfMemberships() a empty onError function to remove the console warning, might not be best fix.


zakgilbert (approved):
Would be a good meeting topic to discuss what should be done if *getNumberOfMemberships()* returns onError.

PR #502, opened 2022-01-27
Removing invalid members to Include and Exclude shows wrong error mod…

zakgilbert:
…al message

**Commits** (Squashed)
- Use dynamic modal for the case of no valid uses were removed.
- Use dynamic modal for the case of last owner being removed.
- Use dynamic modal for the case of last admin being removed.
- Delete *removeErrorModal.html*.
- Fix dialogue.

**Tests**
- [x] Unit Tests
- [x] jasmine Tests

[GROUPINGS-985](https://www.hawaii.edu/jira/browse/GROUPINGS-985)


src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -1412,8 +1403,7 @@
                     listName: "owners"
                 });
             } else {
-                const userType = "owner";
-                $scope.createRemoveErrorModal(userType);
+                $scope.launchDynamicModal(Message.Title.LAST_OWNER_NOT_REMOVED, Message.Body.LAST_OWNER_NOT_REMOVED);
```
zakgilbert, line 1406: I believe we need a Modal controller this Dynamic modal is being used across multiple controllers making it difficult to write tests for.

mhodgesatuh (approved):
Looks fine as far as I can tell.

PR #513, opened 2022-02-02
Fix checkbox UI regression

William-Liang808:
[Groupings - 1018](https://www.hawaii.edu/jira/browse/GROUPINGS-1018)


mhodgesatuh:
style="margin-left: -1.25rem" we generally avoid.  Ideally we would use bootstrap.  If that won't work css (spss).

PR #523, opened 2022-02-11
Fixed broken import button

Rubegen:
First ticket completed with Kobey. Did visual inspection to check and see if it works. Passed Unit test.

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -636,8 +636,8 @@
          * function is called implicitly from include.html and exclude.html.
          * The file is retrieved from the html input with id 'upload'.
          */
-        $scope.readTextFile = function () {
-            let file = input.files[0];
+        $scope.readTextFile = function ($event) {
+            let file = document.getElementById('upload').files[0];
             if (file === undefined) {
                 console.log("undef");
             }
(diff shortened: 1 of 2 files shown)


mhodgesatuh (commented):
let file = document.getElementById('upload').files[0];
There should be an angularJS way to accomplish this.  We try to avoid jQuery and direct DOM references.

Kobeyarai808 (changes requested):
Need to adjust to use Angular, marking "request changes" to clean workflow.

mhodgesatuh (changes requested):
As discussed Friday.  There is an angular way to do this.  Just need to include the right library.  Chak shared a link at the Friday meeting.  Try DM him.

fduckart:
Just dawned on me... instead of adding a third-party library, let's try a compromise that does it the "Angular way," but isolates the warts. 

Add this to the app.js file:
```
UHGroupingsApp.directive('fileModel', [ '$parse', function($parse) {
    return {
        restrict: 'A',
        link: function(scope, element, attrs) {
            var model = $parse(attrs.fileModel);
            var modelSetter = model.assign;

            element.bind('change', function() {
                scope.$apply(function() {
                    modelSetter(scope, element[0].files[0]);
                });
            });
        }
    };
} ]);

```

Then in the input tag for the html page, add the "file-model" attribute. E.g.,
`<input type="file" file-model="somevariablename" ...`

PR #537, opened 2022-03-16
Sanitize Uploaded Files

devgav:
# Ticket Name

Sanitize Uploaded Files

# List of squashed commits

- Implement sanitize function on UI side
- Fix trailing comma issue
-  Remove the .pop

# Test Checklist

- [x] Integration Passed:
- [x] Unit Passed:
- [x] General Inspection

# Link to Ticket

[[Ticket](Insert Link Here)](https://www.hawaii.edu/jira/secure/RapidBoard.jspa?rapidView=100&view=detail&selectedIssue=GROUPINGS-1047)

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -653,6 +653,8 @@
             reader.onload = function (e) {
                 let str = e.target.result;
                 $scope.usersToAdd = (str.split(/[\r\n]+/).join(" ")).slice().split(" ");
+                // Remove trailing comma from the array
+                $scope.usersToAdd.pop();
                 let sanitizedFile = [];
                 for (const users of $scope.usersToAdd) {
                     let sanitizedName = sanitizer(users);
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -627,15 +627,9 @@
 
         // Checks that a users name matches the pattern of either a valid uid or a uhUuid
         function sanitizer(name) {
-            const checkUhuuid = new RegExp("^[a-zA-Z0-9]*$");
-            const checkUid = new RegExp("^[0-9]{8}$");
-            const validUhuid = name.match(checkUhuuid);
-            const validUid = name.match(checkUid);
-            if (validUhuid !== null) { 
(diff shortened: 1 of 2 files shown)


mhodgesatuh:
Note that I've shared the regex expressions with Julio in case he's aware of edge cases that need to be anticipated.  Will update here when I get feedback.

mhodgesatuh:
Julio's code review feedback regarding the regex patterns in the general controller:

There are often exceptions, so I'm worried about putting this in the code. The UH Number for misc usernames is currently the same as the username. There are usernames with leading underscores for application accounts, and those also are treated as the UH Number. There might be exceptions to the username being 8 characters, but I'll have to verify if they are usernames or just an alias in Google. There may be some old usernames with underscores in them.

The code has the regexs swapped too. UhUuid is typically the numeric one of size 8.

Better to sanitize for security (avoid large strings, code injection, etc.) and make some assumptions that are likely to work in the future rather than hardcode the format of the uhUuid/uid in the application code.

Julio's recommendation:

Maybe something simple like alphanumeric and underscores are allowed with a maximum of 30 chars for either uh number or uh username
no spaces, no punctuation, etc. of course

mhodgesatuh (changes requested):
See the comments Julio provided.  The regex patterns, besides being switched, are going to trip on the uhUuid and uid edge cases.  Do not promote this to 220 as is.

devgav:
if it turns out that our definition of a uhuuid and a uid can't be easily covered using regex then I can use an alternative solution, with [DOMPurify](https://www.npmjs.com/package/dompurify), which filters out what dompurify considers harmful code. It will still leave the <></> tags in but remove the harmful part of the tag.

mhodgesatuh:
Actually, the regex solution is very straight forward.  We just need to agree on what characters are allowed, the the max length.  After consultation with the IAM team, the allowed characters are: [A-Z][a-z][0-9]-_ and a length from 8-64 characters.

devgav:
LMK if there are any more valid/invalid inputs that you would like me to add to the in the sanitize test function.

Kobeyarai808 (approved):
Looks good to me.

mhodgesatuh (approved):

mhodgesatuh:
The regex patterns have been corrected, so my concerns have been addressed.

PR #540, opened 2022-03-22
Removes duplicate lines

William-Liang808:


devgav:
Seems fine, good catch removing the duplicate code

PR #544, opened 2022-03-24
MultiAdd feature incorrectly displays info

zakgilbert:
*Commits*
- Remove users from usersToAdd String that are already in group.
---
[GROUPINGS-1052](https://www.hawaii.edu/jira/browse/GROUPINGS-1052)


mhodgesatuh (commented):
What's the 10 in: "uid": data[parseInt(i, 10)].uid,
I'm not finding this to be self explanatory.  I realize that it was there before you started working on it for this ticket.

PR #545, opened 2022-03-25
 MultiAdd feature incorrectly displays info

zakgilbert:
*Commits*
- Add else condition to ***handleSuccuessfulAdd()***.
---
[GROUPINGS-1052](https://www.hawaii.edu/jira/browse/GROUPINGS-1052)


mhodgesatuh (approved):
I don't see any issues with this PR.

PR #565, opened 2022-04-21
Finish up jasmine test and organize on admin.controller.js

William-Liang808:


mhodgesatuh:
Highly recommend you deal with the codacy issues sooner rather than later.  One of which is to use double instead of single quotes.  I saw a lot of those.

PR #569, opened 2022-04-26
Created Regex Global Pattern

Kobeyarai808:
# Ticket Link

[Groupings-1082](https://www.hawaii.edu/jira/browse/GROUPINGS-1082)

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -24,6 +24,7 @@
         $scope.resetUser = [];
         $scope.resetID = [];
         $scope.resetName = [];
+        $scope.uhUuidPatern = "^[_?a-z-?@?0-9]{3,64}$";
 
         // Batch delete
         $scope.memberToRemove = "";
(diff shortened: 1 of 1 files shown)


devgav (approved):

mhodgesatuh (approved):
Note that the pattern works for both the uid and the uhUuid, by design, so it's name is not perfect, but that is already.  However, the inline documentation  above 638 should clarify this.

mhodgesatuh (approved):
Looks good

mhodgesatuh:
Are we still getting the eslint checks, etc?  It is not clear that we are and they are important.

Kobeyarai808:
It looks like Codacy is working on my end. Are you saying that Codacy does not check eslint in it anymore?
Screen Shot 2022-04-29 at 2 13 40 PM

mhodgesatuh:
Thanks for highlighting the codacy call.  I didn't see it.  Glad it is there.

PR #578, opened 2022-05-04
Update Coverage Badge Reference

Erick-Orozco-Ciprian:
Coverage Badge now references the correct script and file

# Ticket Link

[Ticket](https://www.hawaii.edu/jira/browse/GROUPINGS-1068)

# List of squashed commits

- Changed coverage status badge reference 

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ UH Groupings utilizes the Internet2 Grouper project.  Grouper is an enterprise a
 
 [![License](https://img.shields.io/hexpm/l/plug.svg)](https://github.com/apereo/cas/blob/master/LICENSE)
 [![Build Status](https://travis-ci.org/uhawaii-system-its-ti-iam/uh-groupings-ui.png?branch=master)](https://travis-ci.org/uhawaii-system-its-ti-iam/uh-groupings-ui)
-[![Coverage Status](https://https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/blob/badges/jacoco.svg)](https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/blob/master/.github/workflows/coverage.yml)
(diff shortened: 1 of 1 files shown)


mhodgesatuh (approved):
Looking forward to seeing this badge for the project.  What's the plan for the API side of the project?

Erick-Orozco-Ciprian:
> Looking forward to seeing this badge for the project. What's the plan for the API side of the project?

Had to update the directory slightly since It goes through the actions directory and not the file directly. The coverage seems to be updating fine so I will make a PR for the API side aswell!

Kobeyarai808 (approved):
Great job!

mhodgesatuh:
Can the os matrix include centos?  Reason I ask is because it is closer to RHEL, the OS that we use in production.

Erick-Orozco-Ciprian:
> Can the os matrix include centos? Reason I ask is because it is closer to RHEL, the OS that we use in production.

Does not seem to be one of the OSs that GitHub actions have hosted. It does seem possible to set up your own but that requires a bit more work so it may be better to create a separate ticket for the implementation of centos.

mhodgesatuh:
How about fedora?

Probably not necessary to spend too much more time on this.

Michael Hodges
Middleware, Identity and Access Management
University of Hawaii, Information Technology Services
***@***.***
808-956-7195 (GMT -10:00)

we>$

Erick-Orozco-Ciprian:
> How about fedora? Probably not necessary to spend too much more time on this. 

I think github only supports macos, ubuntu, and windows. As mentioned I believe you can create your own "runner" and self-host but it would require a couple of more steps. I'll check in with Frank once these tickets are resolved to see if he has an idea about it!

devgav (approved):

PR #580, opened 2022-05-04
Revised Workflow files 

Erick-Orozco-Ciprian:
# Ticket Link

[Ticket](https://www.hawaii.edu/jira/browse/GROUPINGS-1093)

# List of squashed commits

- Changed to test Java 17 not 18
- Generalized Coverage.yml config

# Test Checklist

- [x] Unit Tests Passed::


mhodgesatuh (approved):
Looks good.  Under what circumstances does the email address get utilized?  Reason I ask is that likely spammers will harvest the email address, so we need to weigh the pros and cons of exposing it.

fduckart:
I think we can make up a random no-reply address since this address is just for the git logging. Maybe something like "actions@noreply.its.hawaii.edu" or some such string.

mhodgesatuh:
Thanks Frank.  Let's run with this suggestion.

Kobeyarai808 (approved):
I like these new changes.

devgav (approved):

PR #607, opened 2022-06-13
Jasmine Test getSyncDestList is failing

zakgilbert:
***COMMITS***
- Match path in groupings.service to path in groupings.service.test.
---
[GROUPINGS-1138](https://www.hawaii.edu/jira/browse/GROUPINGS-1138)
- [x] Unit


mhodgesatuh (commented):
seems like we are stepping back from the RESTful kabob-case recommendation in this instance. ?

zakgilbert:
> seems like we are stepping back from the RESTful kabob-case recommendation in this instance. ?

Good point changing it now.

PR #609, opened 2022-06-16
Create testing/refactor network call in GroupingRestController

zakgilbert:
***COMMITS***
- Refactor RestController URI query building.
- Refactor GroupingsService URI query building.
---
[GROUPINGS-1120](https://www.hawaii.edu/jira/browse/GROUPINGS-1120)
- [x] Unit
- [x] Jasmine
- [x] Visual Inspection

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
@@ -6,6 +6,7 @@ import javax.annotation.PostConstruct;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.jasig.cas.client.util.URIBuilder;
 import org.owasp.html.PolicyFactory;
 import org.owasp.html.Sanitizers;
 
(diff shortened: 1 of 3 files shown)


zakgilbert:
Kinda fumbling with the Codacy errors but still working on it.

fduckart:
Not sure if we have them already, but I would have expected to see a bunch of tests checked in with these kind of changes.

devgav:
> Kinda fumbling with the Codacy errors but am working on it.

With ES6 you should be able to just do this:

` let params = {
       page,
       size,
       sortString,
       isAscending
 };`

https://attacomsian.com/blog/javascript-object-property-shorthand

Pretty sure this will solve some of the codacy errors as well

devgav:
Also, for your query you could do something like this instead of using the Object.Keys

`                let query = (params) => { 
                    let value = [];
                    for (const key in params) {
                        value = encode(key) + "=" + encode(params[key])
                    }
                    value.join("&");
                    return value;
                }`

devgav (approved):

mhodgesatuh (approved):
URIBuilder is a nice addition, especially since it reduces custom code.

PR #615, opened 2022-07-01
Add babel dependency to pom file

devgav:
# Ticket Link

[GROUPINGS-323](https://www.hawaii.edu/jira/browse/GROUPINGS-323)

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,6 +193,12 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.9</version>
         </dependency>
+        <dependency>
+            <groupId>com.jarslab.maven</groupId>
+            <artifactId>babel-maven-plugin</artifactId>
+            <version>1.6</version>
+            <type>maven-plugin</type>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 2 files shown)


mhodgesatuh (commented):
Why do we need the babel dependency?

devgav:
> Why do we need the babel dependency?

Babel mainly helps with browser compatibility by changing new js syntax to js syntax that is supported across older browsers.

mhodgesatuh (approved):
I see no issues

devgav:
Here is a list of browser versions that don't currently support the ES6 syntax -> https://caniuse.com/?search=es6

The browser versions that don't support ES6 is mainly from 2009 - 2013. And the ones that have partial support range from 2014-2017.

mhodgesatuh:
The official supported ITS browsers: <https://www.hawaii.edu/askus/717>
As long as we support these we are good.
 - Michael

fduckart:
Don't you have to do something with that babel dependency for it to be able to work it's magic?

PR #627, opened 2022-07-26
Added which person is being managed in admin

JorWo:
# Ticket Link

[Groupings-813](https://www.hawaii.edu/jira/browse/GROUPINGS-813)

# List of squashed commits

- Moved error div for UH username input under its input box

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/admin.controller.js
+++ b/src/main/resources/static/javascript/mainApp/admin.controller.js
@@ -16,6 +16,7 @@
         $scope.personList = [];
         $scope.pagedItemsPerson = [];
         $scope.currentPagePerson = 0;
+        $scope.currentManagePerson = "";
         $scope.selectedGroupingsPaths = [];
         $scope.emptySelect = false;
         $scope.selectedOwnedGroupingsNames = [];
@@ -67,6 +68,12 @@
                     $scope.searchForUserGroupingInformationOnSuccessCallback,
                     $scope.searchForUserGroupingInformationOnErrorCallback,
                     $scope.personToLookup);
+                groupingsService.getMemberAttributes($scope.personToLookup, function (person) {
+                    $scope.initMemberDisplayName(person);
+                    $scope.currentManagePerson = "(" + $scope.fullName + ", " + $scope.uid + ", " + $scope.uhUuid + ")";
+                }, function () {
+                    $scope.currentManagePerson = "";
+                });
             }
         };
 
--- a/src/main/resources/templates/admin.html
+++ b/src/main/resources/templates/admin.html
@@ -152,9 +152,10 @@
                 </div>
                 <!-- Manage Person -->
                 <div class="tab-pane fade" id="manage-person" role="tabpanel">
-                    <div class="row m-auto pt-3 pb-3">
-                        <div class="col-lg-9 col-md-8 col-12 p-0">
-                            <h2 class="card-title mt-md-1 mt-0 mb-1">Manage Person</h2>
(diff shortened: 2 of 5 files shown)


mhodgesatuh (changes requested):
<div class="mr-auto col-lg-3 col-md-4 col-12 p-0" **style**="overflow:visible"> raises a concern for me since we have a guiding principle to not embed styles in the code.  If we can't make bootstrap meet our needs, then we fall back to creating css code.  Hopefully there aren't other, similar examples in our codebase.

JorWo:
> `div class="mr-auto col-lg-3 col-md-4 col-12 p-0" style="overflow:visible"` raises a concern for me since we have a guiding principle to not embed styles in the code. If we can't make bootstrap meet our needs, then we fall back to creating css code. Hopefully there aren't other, similar examples in our codebase.

Oops I forgot to take that out while playing around with CSS. I just removed it.

mhodgesatuh (approved):
I don't see any issues.

src/main/resources/templates/admin.html
```diff
@@ -198,16 +199,15 @@ <h2 class="card-title mt-md-1 mt-0 mb-1">Manage Person</h2>
...
-
-                            <div ng-show="!errorDismissed2 && !loading && emptySelect"
-                                 class="alert alert-danger form-alert fade show"
-                                 role="alert">
-                                <p th:text="#{screen.message.modal.search.noChecks}"></p>
-                            </div>
-
                         </div>
-
+                        <div class="break"></div>
```
devgav, line 203: If  you intention was to make a break here I would not recommend using a div tag and adding css to it. The more conventional way to add a break is to use the br tag in HTML. If it was not to make a break then i would recommend renaming your css class as it can be confusing to members whether you are making a break or not.

devgav (changes requested):

src/main/resources/templates/admin.html
JorWo: This is a break for pushing all the elements after it to the next line. I tried using <br> but it doesn't work with display:flex elements. I can rename the class to line-break to make it more clear.

JorWo:
I just moved groupingsService.getMemberAttributes() in admin.controller.js to a better location

devgav:
> I just moved groupingsService.getMemberAttributes() in admin.controller.js to a better location

What part of the code did you move the groupingsService.getMemberAttribtues() from on github it appears that you added this groupingsService call in there to display data.

JorWo:
> > I just moved groupingsService.getMemberAttributes() in admin.controller.js to a better location
> 
> What part of the code did you move the groupingsService.getMemberAttribtues() from on github it appears that you added this groupingsService call in there to display data.

Sorry I meant that I added groupingsService.getMemberAttribtues() and moved it from being inside searchForUserGroupingInformationOnSuccessCallback() to searchForUserGroupingInformation()

devgav (approved):

src/main/resources/static/javascript/mainApp/admin.controller.js
```diff
@@ -47,6 +48,12 @@
         $scope.searchForUserGroupingInformationOnSuccessCallback = function (res) {
             $scope.personList = _.sortBy(res, "name");
             $scope.filter($scope.personList, "pagedItemsPerson", "currentPagePerson", $scope.personQuery, true);
+            groupingsService.getMemberAttributes($scope.personToLookup, function (person) {
```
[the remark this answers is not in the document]
JorWo: Got it

PR #636, opened 2022-08-02
Gets rid of unused properties

carolwong492:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-735)

# List of squashed commits
- Re-adds emptying table

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -23,74 +23,31 @@ cas.send.renew=false
 # =========================================================================
 # Groupings api
 groupings.api.settings=uh-settings
-groupings.api.attribute_assign_id_size=50
-groupings.api.grouping_admins=${groupings.api.settings}:groupingAdmins
-groupings.api.grouping_apps=${groupings.api.settings}:groupingApps
-groupings.api.grouping_owners=${groupings.api.settings}:groupingOwners
-groupings.api.grouping_superusers=${groupings.api.settings}:groupingSuperusers
 groupings.api.attributes=${groupings.api.settings}:attributes
 groupings.api.for_groups=${groupings.api.attributes}:for-groups
-groupings.api.for_memberships=${groupings.api.attributes}:for-memberships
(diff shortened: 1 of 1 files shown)


mhodgesatuh (approved):
I don't see any issues

src/main/resources/application.properties
```diff
@@ -23,75 +23,32 @@ cas.send.renew=false
...
 groupings.api.include=:include
 groupings.api.owners=:owners
-groupings.api.assign_type_group=group
-groupings.api.assign_type_immediate_membership=imm_mem
-groupings.api.subject_attribute_name_uuid=uid
-groupings.api.operation_assign_attribute=assign_attr
-groupings.api.operation_remove_attribute=remove_attr
-groupings.api.operation_replace_values=replace_values
-groupings.api.privilege_opt_out=optout
-groupings.api.privilege_opt_in=optin
```
JorWo, line 58: These two lines are still needed in GroupAttributeServiceImpl.java

devgav: I believe the isTrio is needed in there as well. Good catch

devgav:
@carolwong492 i believe that intelliJ has trouble distinguishing unused variables in our application.properties file. Because of this you likely need to do a global search for a lot of these variables which I personally have had some trouble with in the past using intelliJ, instead I use VSCode since the search feature is better.

src/main/resources/application.properties
carolwong492: Are you talking about the UI or the API? This pull request is for the UI, not the API. I don't think that file exists on the UI side.

devgav: > This pull request is for the UI, not the API. I don't think that file exists on the UI side.

oops, sorry 2about that @carolwong49 seems like there was a mix up on the branch that this pr was going into. Well at least now you will know to not remove those from the API haha.

devgav (approved):

PR #647, opened 2022-08-12
Fixes missing error message for empty input when removing in include

KaileeHung:
# Ticket Link

[GROUPINGS-1180](https://uhawaii.atlassian.net/browse/GROUPINGS-1180)

# List of squashed commits

- Fixes missing error message for empty input when removing in include
- Removes console error when removing invalid user
- Adds Jasmine testing and adjusts error message for remove
- Adds functions for on-click events
- Fixes Codacy errors
- Creates dismissErrors function

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:


mhodgesatuh (commented):
$scope.memberToRemove = membersToRemove;
$scope.userInput = $scope.memberToRemove;

overall looks good I think.  The above seems a bit odd, like new code is being bolted onto old.  See the above.

fduckart:
I'd be super great if you could figure out a way to move those compound statements for the on-click events in the html page you have going now, and move them into a javascript function... then we'd have the logic in the controller, for better or worse... but hopefully better.

devgav:
Seems like there are some failing tests particularly with java 11.... Could be something on github actions i have been noticing some bugs with github actions lately. I would suggest rerunning our github actions and see if you get a full pass.

devgav (approved):
Looks good. Thanks for adding in the unit tests.

Out of curiosity is the error message missing on the exclude remove feature as well?

KaileeHung:
> Looks good. Thanks for adding in the unit tests.
> 
> Out of curiosity is the error message missing on the exclude remove feature as well?

No, it was only missing in the Include for some reason

PR #664, opened 2022-09-29
[UI Side] Changes add/remove member request to contain a body

JorWo:
# Ticket Link

[Groupings-1214](https://uhawaii.atlassian.net/browse/GROUPINGS-1214)

# List of squashed commits

- Changed all add and remove member functions in general.controller.js to make a PUT request with an ArrayList instead of a comma-separated string of members
- Added makeApiRequestWithBody and makeApiRequestWithBodyAndTimeoutModal to groupings.service.js
- Changed GroupingsRestController to expect a PUT request with a body instead of a header
- Added sanitizeList helper method with 100% Jacoco test coverage to GroupingsRestController to sanitize a list of strings
- Added another makeApiRequestBody in HttpRequestService that takes a list of data instead of string in the parameter
- Added JsonUtil and JsonUtilTest from the API side to use in unit tests (Convert list into json when performing mockMvc.content())
- Updated Unit and Jasmine tests

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
@@ -1,7 +1,9 @@
 package edu.hawaii.its.api.controller;
 
 import java.security.Principal;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import javax.annotation.PostConstruct;
@@ -89,7 +91,7 @@ public class GroupingsRestController {
 
     // Constructor.
     public GroupingsRestController() {
-        policy = Sanitizers.FORMATTING.and(Sanitizers.LINKS);
+        policy = Sanitizers.FORMATTING;
     }
 
     /*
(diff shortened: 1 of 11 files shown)


mhodgesatuh:
Codacy has issues with the PR.  I note that for the injection sync it doesn't have advice.  Found this, <https://stackoverflow.com/questions/51715616/how-to-fix-codacy-alert-generic-object-injection-sink>, which reminds me of a conversation we had a while back, so it might not be the best of advice.

JorWo:
> Codacy has issues with the PR. I note that for the injection sync it doesn't have advice. Found this, https://stackoverflow.com/questions/51715616/how-to-fix-codacy-alert-generic-object-injection-sink, which reminds me of a conversation we had a while back, so it might not be the best of advice.

I think we should disable the generic object injection sink because it gives an error anywhere a variable is passed into square brackets. I found that the author of the rule said himself that the rule is too noisy and it was designed to assist with manual checks: https://github.com/nodesecurity/eslint-plugin-security/issues/21#issuecomment-1157887653

mhodgesatuh:
I'm fine with disabling this check.  If you have time, can you locate it for me?  Otherwise it might be next week before I get to it. This week is a very busy one from hereon.  There are other Codacy issues though that should be addressed.

JorWo:
> I'm fine with disabling this check. If you have time, can you locate it for me? Otherwise it might be next week before I get to it. This week is a very busy one from hereon. There are other Codacy issues though that should be addressed.

Here is the link, the rule name is "Detect object injection": https://app.codacy.com/p/383092/patterns/list?engine=f8b29663-2cb2-498d-b923-a10c6a8c05cd&category=2&status=true

mhodgesatuh (commented):
Not sure why the following.  Why is removing 1 handled differently than 2+?  Necessary?  If so the inline comments should explain since not intuitive.

`                $scope.membersToModify = [];
                if (membersToRemove.length > 1) {
                    removeMembers(membersToRemove, listName);
                } else {
                    $scope.userInput = membersToRemove[0];`

mhodgesatuh:
Given that Codacy doesn't feel like this check meets their standards, maybe to ignore the particular warning for now until we know better what the issue is and how to address it.  I'm fine with that strategy for now.

devgav:
I think we should try that for...of loop since it seems like codacy is just complaining about accessing the elements in our array with the brackets and our index. The for...of loop essentially does the same thing without needing the index or the brackets and needing to access our elements with [i].

devgav (approved):
Perfect!

PR #665, opened 2022-10-07
Removes unneeded text in multi-remove modal when all users are valid

KaileeHung:
# Ticket Link

[GROUPINGS-1228](https://uhawaii.atlassian.net/browse/GROUPINGS-1228)

# List of squashed commits

- Removes unneeded text in multi-remove modal when all users are valid

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:


mhodgesatuh (approved):
The ticket doesn't provide screen shots to illustrate the before and after, which would have been useful.

KaileeHung:
> The ticket doesn't provide screen shots to illustrate the before and after, which would have been useful.

Good point! I'll add that in!

devgav (approved):

PR #676, opened 2022-11-11
Fixed displaying complete list of memberships

carolwong492:
# Ticket Link

[GROUPINGS-1275](https://uhawaii.atlassian.net/browse/GROUPINGS-1275)

# List of squashed commits

- Fixed small error

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:


devgav (approved):
Thanks for fixing this so quickly. On our master branch tho we should be just calling the numberOfMemberships endpoint that we have instead of doing this. It's a waste of computation and time on our end since it looks like we are requesting the entire list of memberships then filtering when we only need the number (which we have the function already).

PR #677, opened 2022-11-15
Groupings 852 clean up syntax error in js code

goodroach:
# Ticket Link

[Groupings 852](https://uhawaii.atlassian.net/browse/GROUPINGS-852)

# List of squashed commits

- syntax fix
- used arrow function vs typing it out

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:


mhodgesatuh (commented):
Not sure what happened to create so many changes, but this appears excessive given the goal cleanup js code.  For example, why are there changes to the properties file(s)?

May need to pull out the intended changes and apply them to fresh copy of the master branch, but let's see what other reviewers say.

mhodgesatuh (commented):
Given the number of changes I suspect a high probability of at least a few regressions being introduced.

fduckart:
Something went way wrong, obviously.

devgav (commented):
Something definitely does not look right here. Looks like this pr needs a little bit of help before we can review it further if need help you can message me on slack and we can try figure out what is going on. I think the cleanest and fastest way to solve this is to probably make a new branch off of the master branch tho and apply the changes you made to the code there then push up another pr.

devgav:
Also regarding that ticket about the syntax error. I believe it is a personal preference, but both methods of writing the IIFE is valid and will work. If we want to continue moving forward with this ticket you should fix these changes. If we decide it isn't worth our time for these changes I would recommend leaving a comment in the ticket and closing the JIRA ticket. 

Personally I would say it isn't worth our time to make that change, but I will see what the other branch leads say.

fduckart:
I think we decide to go with very latest syntax, mostly to avoid ambiguity going forward. The work appears to have been done on an old branch or some other merge problem occurred. So, let's fix the PR so it has just the desired javascript tweaks all in a single commit.

devgav:
Closing this PR so Hugh can get a chance to fix this branch.

PR #688, opened 2022-12-21
Changes home page to greet people by first name

carolwong492:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1284)

# List of squashed commits

- Added to tests

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/groupings/access/User.java
+++ b/src/main/java/edu/hawaii/its/groupings/access/User.java
@@ -45,6 +45,10 @@ public class User extends org.springframework.security.core.userdetails.User {
         return attributes.getValue("cn");
     }
 
+    public String getGivenName() {
+        return attributes.getValue("givenName");
+    }
+
     public boolean hasRole(Role role) {
         return getAuthorities().contains(new SimpleGrantedAuthority(role.longName()));
     }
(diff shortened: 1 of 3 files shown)


JorWo (approved):
Looks good! Thanks for adding the tests too.

src/main/java/edu/hawaii/its/groupings/access/User.java
```diff
@@ -45,6 +45,10 @@ public String getName() {
         return attributes.getValue("cn");
     }
 
+    public String getGivenName() {
+        return attributes.getValue("givenName");
+    }
+
```
JorWo, line 51: Does getName() in User return the person's full name and getGivenName() return the first name? If so, can you change getGivenName() to getFirstName() to make it more clear.

fduckart: This one is on me. I was torn about this decision and suggested to Carol that we match up the getter with the underlying attribute in LDAP. I'm not sure what got into the LDAP people to name the attribute the way they did. Whatever you decide is good, I think.

fduckart: P.S. A lot of people (but not all) in the programming world have decided that firstname and/or lastname are considered one word, so need for uppercase on the letter N. If you go in for that line of thinking, of course.

JorWo: In that case, I'll keep it as is.

PR #692, opened 2022-12-27
Add jasmine test and checks for file

goodroach:
# Ticket Link

[1212](https://uhawaii.atlassian.net/browse/GROUPINGS-1212)

# List of squashed commits

- Add jasmine test and checks for file

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -658,6 +658,34 @@
          * The file is retrieved from the html input with id 'upload'.
          */
         $scope.readTextFile = function (inputFile) {
+            if (!$scope.verifyImportFileType(inputFile)) {
+                $scope.launchDynamicModal(
+                    Message.Title.INVALID_FILE,
+                    "File must be a .txt file."
+                );
+                return;
+            }
+            if (!$scope.verifyImportFileSize(inputFile)) {
+                $scope.launchDynamicModal(
+                    Message.Title.INVALID_FILE,
+                    "File is too large. Maximum file size: 5MB"
+                );
+                return;
+            }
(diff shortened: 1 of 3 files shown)


JorWo (approved):
Looks very good! I like the dynamic modals that launch to display file warnings.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -658,6 +658,35 @@
...
+                );
+                return;
+            }
+            if (!$scope.verifyImportFileName(inputFile)){
+                $scope.launchDynamicModal(
+                    Message.Title.INVALID_FILE,
+                    "File name has illegal characters."
+                );
+                return;
+            }
```
JorWo, line 689: Can you use change this to a long else-if chain to avoid using a return statement? Could you also move the all invalid file messages into the app.constants.js file too?

JorWo (changes requested):

src/main/resources/static/javascript/mainApp/general.controller.js
fduckart: Let's just keep this style. The const for the type at the top doesn't appear to be used, though.

mhodgesatuh: Thanks Frank.  I was going to mention this too.  Jorden, academics used to teach that methods should have one exit, at the bottom.  But the craft has evolved.  Guards were introduced.  Avoiding complexity by exiting early has become an accepted practice.

mhodgesatuh (commented):
Codacy reports a syntax issue that is easily fixed.  Also a code duplication issue.

JorWo (approved):

PR #694, opened 2022-12-30
Fix logout position and wrapping. Selected tab cirle now a cirlce

yertsti:
# Ticket Link

[1301](https://uhawaii.atlassian.net/browse/GROUPINGS-1301)

# List of squashed commits

- Fix logout position and wrapping. Selected tab cirle now a cirlce
- Made description text wrap
- Fix positioning of tooltips for the trashcan in include/exclude tab
- Added borders to members table and removed wrapping from listing column
- Added radius to border around members
- Fixed update description of grouping input field
- Moved border to card body, so the border is correct on smaller resolutions

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:


JorWo (approved):
Thanks for fixing up the UI,

JorWo:
There seems to be a merge conflict. Rebasing this branch with master then running the scss compile command should do the trick.

src/main/resources/static/scss/sections/_groupings.scss
```diff
@@ -142,3 +142,8 @@
         }
     }
 }
+
+.form-control {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+}
```
JorWo, line 149: Can you check if this did not affect the other input boxes such as 'Filter Members' inside /manage-groupings and 'Filter Groupings' in /admin /membership /grouping pages?

yertsti: It did. They don't have radius on the right corners.
image
Would it be alright to add an id to the edit description input box and apply the css to that?

JorWo: In selected-grouping.html, there is a rounded-left Bootstrap class around line 55 that does not seem to work. If changing the Bootstrap does not seem to work, add a class instead of an id with a class name similar to 'description-form-button'.

PR #697, opened 2023-01-10
Restructure pagination on the UI-side to be iterative instead of recu…

zakgilbert:
…rsive

# Ticket Link

[GROUPINGS-304](https://uhawaii.atlassian.net/browse/GROUPINGS-304)

# List of squashed commits
- Implement iterative getGrouping().
- Implement new getGrouping() to deal with UI loaders and such.
- Remove old getGroupingInformation().
- Provide jasmine coverage for new fetchGrouping().
- Fix Codacy Errors.
- Add and remove causes list to reset.

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/groupings.service.js
+++ b/src/main/resources/static/javascript/mainApp/groupings.service.js
@@ -8,7 +8,7 @@
      * NOTE: All API requests defined here require at least two parameters, namely onSuccess and onError, which are
      * controller handlers(callback functions) for manipulating data on a successful or unsuccessful request.
      */
-    UHGroupingsApp.factory("groupingsService", function (dataProvider, BASE_URL) {
+    UHGroupingsApp.factory("groupingsService", function (dataProvider, BASE_URL, PAGE_SIZE) {
         return {
 
             /**
@@ -29,7 +29,7 @@
              * @param onSuccess - Function to be called if HTTP request returns as a success.
              * @param onError - Function to be called if HTTP request returns an error.
              */
-            getGrouping(path, page, size, sortString, isAscending, onSuccess, onError) {
+            getGrouping(path, page,size, sortString, isAscending, onSuccess, onError) {
                 let endpoint = BASE_URL + "groupings/" + path + "?";
                 let params = { page, size, sortString, isAscending };
                 let query = this.encodeParameterizedQueryString(params);
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -132,26 +132,23 @@
          * Get the username of the current user
          */
         groupingsService.getCurrentUser((res) => {
-                $scope.currentUser = res.data.username;
-            }
-        );
+            $scope.currentUser = res.data.username;
+        });
 
         /**
          * Get the number of memberships that the current user is associated with.
          */
         groupingsService.getNumberOfMemberships((res) => {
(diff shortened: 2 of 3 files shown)


mhodgesatuh (commented):
Codacy has some suggestions for improving the source code.
The scope of change does not appear to be matched by the new tests, but I'm not sure if that's because they augment the old tests such that code coverage is complete.

zakgilbert:
@mhodgesatuh Fixed the codacy issues and yes, one of the functions getGroupingInformation is rewritten and thus augments the old test of getGroupingInformation. fetchGrouping is a new function and is covered in the jasmine tests.

mhodgesatuh:
Codacy still report a complexity issue.  I notice the constant PAGE_SIZE is passed as a parameter?  Is the method called with values other than this constant?  If not the constant shouldn't be a parameter, and the method should be able to directly reference the static constant from its class.

mhodgesatuh (approved):

src/main/resources/static/javascript/mainApp/groupings.service.js
```diff
@@ -8,7 +8,7 @@
      * NOTE: All API requests defined here require at least two parameters, namely onSuccess and onError, which are
      * controller handlers(callback functions) for manipulating data on a successful or unsuccessful request.
      */
-    UHGroupingsApp.factory("groupingsService", function (dataProvider, BASE_URL) {
+    UHGroupingsApp.factory("groupingsService", function (dataProvider, BASE_URL, PAGE_SIZE) {
```
JorWo, line 11: Don't forget to remove PAGE_SIZE from here.

JorWo:
For the Codacy issue on complexity issue, I think we can raise the complexity from 6 to 10 to accommodate this PR and a few of other functions that are currently labelled with the same issue such as $scope.setPage(). For our Jasmine tests, our current tests that are already written do not provide a whole lot of code coverage. @zakgilbert, I can send you code through Slack when I have the time on how to mock HTTP requests in Jasmine. I recently figured it out in a ticket I have been working on.

mhodgesatuh:
Passing the constant BASE_URL is another red flag.  Unless the method is
known to handle alternative URLs, the argument should be removed and
referenced via static class constant instead.

Raising the complexity might hide some of these and at this point I'm
wondering if there will be more.

 - Michael

zakgilbert:
Right now, and for a long time now, BASE_URL is defined in app.constants.js
as...
UHGroupingsApp.constant("BASE_URL", "api/groupings/");
The variable is used in timeout.controller.js and groupings.service.js and
could definitely be referenced via static class constant. Shall I make a
ticket for this?

mhodgesatuh:
I'm thinking in terms of static JAVA class constants.  Sounds like the
context is jQuery, in which case my comment is not relevant.  And perhaps
we do need to adjust complexity after all.

zakgilbert:
I was able to get Codacy to pass by moving the logic performed in the giant if statement to the API, so now the Class Grouping object in the API contains an isEmpty which checks if all groups are empty(this means that the pagination is complete.)

mhodgesatuh:
Keeping the UI code as simple as possible, where reasonable, is one of our
guiding principles.  Thanks for continuing to investigate the alternatives
and finding a solution.

zakgilbert:
There's still a lot of work to be done on this, thanks for your input, I'll be closing this PR for now.

PR #707, opened 2023-02-03
Icon for last grouping owner

goodroach:
# Ticket Link

[Groupings 1292](https://uhawaii.atlassian.net/browse/GROUPINGS-1292)

# List of squashed commits

- Icon for last grouping owner

# Test Checklist

- [Y] Unit Tests Passed:
- [Y] Jasmine Tests Passed:
- [Y] General Visual Inspection:

### DIFF
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -127,6 +127,7 @@ screen.message.common.tooltip.nav.preferences=Manage this grouping's preferences
 screen.message.common.tooltip.remove.include=Remove this member from the Include members list
 screen.message.common.tooltip.remove.exclude=Remove this member from the Exclude members list
 screen.message.common.tooltip.remove.owner=Remove this owner from the owners list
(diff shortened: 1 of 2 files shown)


mhodgesatuh:
See slack message on suggested text for the tooltip.

mhodgesatuh (commented):
ng-if="groupingOwners.length === 1">

I suspect that this should be == given that this is about count rather than an object equivalency.  Any use of an inequality would further suggest that I'm correct.

JorWo:
> ng-if="groupingOwners.length === 1">
> 
> I suspect that this should be == given that this is about count rather than an object equivalency. Any use of an inequality would further suggest that I'm correct.

It's a good practice to use === instead of == in Javascript since it is not a strongly-typed language. Using == will convert the type, meaning 1 == "1" is true while 1 === "1" is false.  Also, Codacy likes throwing an error for using == instead of ===.

JorWo (commented):
Can you show a picture? This can be done during the meeting too.

PR #708, opened 2023-02-03
Change preferences to synchronization destinations

yertsti:
# Ticket Link

[1321](https://uhawaii.atlassian.net/browse/GROUPINGS-1321)

# List of squashed commits

- Change preferences to synchronization destinations
- Add html attribute to fix tooltip on preferences
- Remove tooltip directive

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -94,7 +94,7 @@ screen.message.modal.empty.warning=The selected group is empty.
 # Preferences
 screen.message.modal.ldap.confirmation=Click Ok to update the CAS/LDAP preference as requested.
 screen.message.modal.emaillist.confirmation=Click Ok to update the Email list preference as requested.
-screen.message.modal.preferences.warning=Please be thoughtful about any changes here as some changes are operationally expensive. Note that in extreme cases a requested change may take several hours.
+screen.message.modal.preferences.warning=Please be thoughtful about any changes here as some changes are operationally very expensive.  Avoid rapidly enabling and disabling a synchronization destination.
 screen.message.modal.remove.error=You are unable to remove this {{ userType }}. There must be at least one {{ userType }} remaining.
 screen.message.modal.soleowner.remove.error=The user you are trying to remove is the sole owner of the following groupings {{ soleOwnerGroupingNames }}.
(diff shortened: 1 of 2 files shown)


JorWo (commented):
I noticed you created a custom directive for the sync destination tooltips. I know that those tooltips have been broken but instead of creating a custom directive, could you try adding this to the HTML?: onmouseover="$(this).tooltip('toggle')". Also, if you could find a better way than adding that HTML attribute, that would be even better.

yertsti:
Ok I'll look into it

yertsti:
I did some googling and if you use uib-tooltip="tooltip" it displays the bootstrap styled tooltip in the ng-repeat block, but the small arrow between the tooltip and the icon doesn't show. I think we would need to update angular or bootstrap or something to get the arrow to show or add some custom css. Also, all the other tooltips in ng-repeat blocks use the onmouseover attribute. Not sure if it is worth trying to figure out how to get uib-tooltip to work.

mhodgesatuh:
Sounds like the uib-tooltip discussion would be a good one for the team meeting.

JorWo:
> I did some googling and if you use uib-tooltip="tooltip" it displays the bootstrap styled tooltip in the ng-repeat block, but the small arrow between the tooltip and the icon doesn't show. I think we would need to update angular or bootstrap or something to get the arrow to show or add some custom css. Also, all the other tooltips in ng-repeat blocks use the onmouseover attribute. Not sure if it is worth trying to figure out how to get uib-tooltip to work.

I tried to get the uib-tooltip to work before but it involves downgrading our Boostrap version, which will break the rest of our UI. The AngularUI Bootstrap library is no longer supported, so I do not expect that bug to be fixed. 

On the bright side, Kristyn is working on a custom directive of a tooltip that mimics the uib-tooltip in this PR: https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/pull/713. You can leave the tooltip with the 'onmouseover' attribute and she will later update the tooltip with our new custom directive

JorWo (approved):

PR #744, opened 2023-03-04
Deletes interface files

Kristyn-Mimura:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1350)

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:


mhodgesatuh (approved):
I don't see any issues

JorWo (approved):
Looks great

JorWo (changes requested):
I think UhCasAttributes would be a better name than UhAttributes because it is more specific.

mhodgesatuh:
That occurred to me too, but I wasn't clear if the attributes were strictly from CAS or if CAS was a subset.  The other thing that occurred to me that is some day we might switch to Shibboleth and drop CAS, and it would be nice not to have to change the variable name.

JorWo:
> That occurred to me too, but I wasn't clear if the attributes were strictly from CAS or if CAS was a subset. The other thing that occurred to me that is some day we might switch to Shibboleth and drop CAS, and it would be nice not to have to change the variable name.

That seems better to not have to change the name. I'll proceed with merging.

JorWo (approved):

PR #748, opened 2023-03-09
Enables enter key usage for checkboxes

Kristyn-Mimura:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1383)

# List of squashed commits
- Implements enter key functionality to exclude.html & include.html
- renames createSyncDestModal
- Uses html for enter key usage for remaining checkboxes
- Updates to use angularJS for all enter-key functionality for checkboxes 

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -1777,7 +1777,7 @@
         };
 
         $scope.updateIncludeCheck = function () {
-            $scope.includeCheck = $scope.includeCheck === false;
+            $scope.includeCheck = !$scope.includeCheck;
         };
 
         $scope.updateExcludeCheck = function () {
@@ -2099,6 +2099,20 @@
             }
             return "/uhgroupings";
         };
+
+        $scope.clickCheckBox = function (i) {
+            if (event.keyCode === 13) {
+                $scope.membersInCheckboxList[i.uhUuid] = !$scope.membersInCheckboxList[i.uhUuid];
+            }
+        };
+
+        $scope.clickWithEnter = function(event) {
+            if (event.keyCode === 13) {
+                event.target.click();
+                event.preventDefault();
+            }
+        };
+
     }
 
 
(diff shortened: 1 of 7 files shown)


JorWo:
For all places you used onkeydown, could you use a ternary operator like this: 
`onkeydown="event.keyCode === 13 ? this.click() : null"`. 
This is to match the other places in the HTML that use event.keyCode === 13. 

Also, could you add using the enter key to the rest of the checkboxes such as in preferences as well as the checkboxes for every member in include/exclude?

One more thing, although it is not in the scope of the ticket, we can fix it now. I noticed in sync-destinations.html that we forgot to rename createSyncDestModal to displaySyncDestModal. So currently our sync destinations checkbox is not working properly. Thanks!

JorWo:
What's the status of this PR?

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -1777,7 +1777,8 @@
         };
 
         $scope.updateIncludeCheck = function () {
-            $scope.includeCheck = $scope.includeCheck === false;
+            //$scope.includeCheck = $scope.includeCheck === false;
```
JorWo, line 1780: There is a leftover comment here.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -2099,6 +2099,18 @@
             }
             return "/uhgroupings";
         };
+
+        $scope.clickCheckBox = function (i) {
+            $scope.membersInCheckboxList[i.uhUuid] = !$scope.membersInCheckboxList[i.uhUuid];
+        };
```
JorWo, line 2105: Can you move the `if (event.keyCode === 13 )` condition into this function just like you did with `$scope.clickWithEnter()`? Then in places in HTML that have: `ng-keydown="$event.keyCode === 13 ? clickCheckBox(i): null"`, you can get rid of `$event.keyCode === 13`.

Kristyn-Mimura: I made the changes.

JorWo (approved):

PR #766, opened 2023-03-24
Add host name to feedback email

yertsti:
# Ticket Link

[882](https://uhawaii.atlassian.net/browse/GROUPINGS-882)

# List of squashed commits

- Add host name to feedback email

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/groupings/service/EmailService.java
+++ b/src/main/java/edu/hawaii/its/groupings/service/EmailService.java
@@ -38,12 +38,25 @@ public class EmailService {
 
     public void send(Feedback feedback) {
         logger.info("Feedback received in EmailService: " + feedback);
+
+        InetAddress ip;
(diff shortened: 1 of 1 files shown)


mhodgesatuh (approved):
I don't see any issues.

JorWo (commented):
Is it possible to avoid using a try-catch block?

yertsti:
Would it be better to have the method throw the exception instead of a try-catch block?

JorWo:
> Would it be better to have the method throw the exception instead of a try-catch block?



> Would it be better to have the method throw the exception instead of a try-catch block?

Sorry, I just saw that EmailService runs the same way with a try-catch. We can keep the try-catch for now.

JorWo (approved):

PR #789, opened 2023-04-19
Override recipient email

yertsti:
# Ticket Link

[1334](https://uhawaii.atlassian.net/browse/GROUPINGS-1334)

# List of squashed commits

- Override recipient email
- Add override to email recipient
- Fix email unit test
- Add property to properties skeleton
- Add space

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/test/java/edu/hawaii/its/groupings/service/EmailServiceTest.java
+++ b/src/test/java/edu/hawaii/its/groupings/service/EmailServiceTest.java
@@ -5,6 +5,7 @@ import edu.hawaii.its.groupings.type.Feedback;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.mail.MailException;
 import org.springframework.mail.MailSendException;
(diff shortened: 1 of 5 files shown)


mhodgesatuh (commented):
to is an odd variable name.  Naming it for its contents is a better practice, which means it will be a noun.

src/test/java/edu/hawaii/its/groupings/service/EmailServiceTest.java
```diff
@@ -49,6 +49,7 @@ public void send(SimpleMailMessage mailMessage) throws MailException {
 
         emailService = new EmailService(sender);
         emailService.setEnabled(true);
+        emailService.setTo("address");
```
JorWo, line 52: Is creating the setter `setTo` needed for this unit test? Could you check the JaCoCo test coverage for this test with and without the setter? Here's the documentation: https://uhawaii.atlassian.net/wiki/spaces/SITARd/pages/12943618/jaCoCo+Code+Coverage

yertsti: Without the setter, I would have to add a null check in the send methods of EmailService when it checks if the recipient email != the team's email list. Below are screenshots of the Jacoco reports.
Main: 
image

No setTo
image

setTo
image

JorWo: Could you try to get the test coverage to 100% or as high as you can? Thanks.

yertsti: image
I couldn't figure out how to force a unit test down the path where the InetAddress.getLocalHost() throws an UnknownHostException

yertsti:
> to is an odd variable name. Naming it for its contents is a better practice, which means it will be a noun.

I renamed it to email.send.recipient. I think it sounds better.

src/test/java/edu/hawaii/its/groupings/service/EmailServiceTest.java
JorWo: Looks great. Also, could you switch the Hamcrest asserts into Junit assertions? So instead of using assertThat from Hamcrest, we use assertEquals. Note that in assertEquals, the first parameter is the expected value.

JorWo (approved):
Good work! Thanks for getting the code coverage up to 100%.

mhodgesatuh (approved):
Changing "to" to "recipient" is a nice catch.  The more ambiguity we remove the better.

PR #803, opened 2023-05-11
Fix success modal for reset grouping

yertsti:
Add debug message to catches

# Ticket Link

[1454](https://uhawaii.atlassian.net/browse/GROUPINGS-1454)

# List of squashed commits

- Fix success modal for reset grouping
- Add debug message to catches

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/static/javascript/mainApp/general.controller.js
+++ b/src/main/resources/static/javascript/mainApp/general.controller.js
@@ -18,7 +18,6 @@
         $scope.invalidMembers = [];
         $scope.membersInList = "";
         $scope.isMultiAdd = false;
-
         // Remove members
         $scope.multiRemoveResults = [];
         $scope.membersToRemove = [];
@@ -892,7 +891,7 @@
                     $scope.getGroupingInformation();
                     $scope.syncDestArray = [];
                 }
-            });
+            }).catch(() => {});
         }
 
         /**
@@ -1307,7 +1306,7 @@
                 $scope.loading = true;
                 clearMemberInput();
                 $scope.getGroupingInformation();
-            });
+            }).catch(() => {});
         }
 
         /**
@@ -1489,6 +1488,7 @@
                     }
                 }
             }
+
             // Otherwise, display the result success modal.
             $scope.displaySuccessfulGroupResetModal((() => {
                 if (results.includeSuccess && results.excludeSuccess) {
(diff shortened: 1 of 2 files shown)


JorWo (commented):
What are the pros of using $log.debug?

mhodgesatuh:
In production the amount of information logged should be minimal, but
sufficient to help with initial troubleshooting.  Reason for this is that
in most production environments there are a huge number of users and the
logs can fill up disk space quickly if not thoughtfully populated.
Services can fail when they run out of disk space.  Note that I have
created a ticket defining how best to deploy log levels for our project.

yertsti:
$log.debug is there to not have an empty catch for the rejected promise when the success modals are dismissed instead of closed. It doesn't matter if the promise is rejected, so it probably could be an empty catch.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -910,7 +909,7 @@
                     $scope.getGroupingInformation();
                     $scope.syncDestArray = [];
                 }
-            });
+            }).catch(() => { $log.debug("Modal dismissed"); });
```
JorWo, line 912: I remember you do not need to use .catch for these .result.finally() calls. The .result.finally calls takes in two functions as parameters, one for on success and another for on error. So you can change this exact line to:
`}, () => $log.debug("Modal dismissed"));` or just `}, $log.debug("Modal dismissed"));`, I'm not completely sure if the second one is allowed.

yertsti: I tried result.finally() with two functions, but the second one isn't called when the modal is dismissed and the promise is rejected. It also still prints an error in console about a possible unhandled exception. 
It worked for result.then() with two functions, but then we would have to duplicate the stuff in the success function into the rejected function. Also, should I remove $log.debug stuff cause I don't think we care if the modal is dismissed or not for the success modals.

JorWo: Yes, I agree with removing $log.debug. We can go with an empty .catch() block. Other places in the general controller also use an empty catch block.

JorWo (approved):

JorWo:
Can you backport this to devrc-2-3-0?

yertsti:
Made the PR for devrc-2-3-0

PR #806, opened 2023-05-19
Add CSV option for batch imports

yertsti:
Add example for CSV file in import modal

Update import modal

# Ticket Link

[1457](https://uhawaii.atlassian.net/browse/GROUPINGS-1457)

# List of squashed commits

- Add CSV option for batch imports
- Add example for CSV file in import modal
- Update import modal

# Test Checklist

- [x] Unit Tests Passed:
- [x] Jasmine Tests Passed:


mhodgesatuh (commented):
The constant "uh number" should be a constant and referenced in both where it is used to create the output header row and where it is referenced to detect the presence of a header row.

yertsti:
I made a constant named UUID_COLUMN_HEADER in the csv constants.

mhodgesatuh (approved):
Thanks for ensuring that all of the headers are consistently titlecase.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -650,8 +650,24 @@
...
+                    const indexOfUhNumber = firstRow.findIndex((header) => header.includes(Message.Csv.UUID_COLUMN_HEADER));
+                    if (indexOfUhNumber < 0) {
+                        $scope.displayDynamicModal(
+                            Message.Title.INVALID_FILE,
+                            `CSV file does not contain ${Message.Csv.UUID_COLUMN_HEADER} in the header`,
+                        );
+                        return;
+                    }
+
+                    const UHNumbersInFile = namesInFile.map((row) => row.split(",")[parseInt(indexOfUhNumber, 10)]).slice(1, -1);
```
JorWo, line 665: I see you used parseInt() to get around the Codacy issue. I recommend using Number() instead because it is a little cleaner to read.

src/main/resources/static/javascript/mainApp/general.controller.js
```diff
@@ -824,8 +841,8 @@
             $scope.waitingForImportResponse = true; // Small spinner on
             groupingsService.invalidUhIdentifiers(uhIdentifiers, (res) => { // Check for invalid uhIdentifiers
                 $scope.waitingForImportResponse = false; // Small spinner off
-                $scope.isBatchImport = uhIdentifiers.length > Threshold.MULTI_ADD;
-
+                $scope.isBatchImport = angular.element(document.querySelector("#upload"));
+                $scope.isLargeBatchImport = uhIdentifiers.length > Threshold.MULTI_ADD;
```
JorWo, line 845: Is there a way to avoid having to separate a batch import and a large batch import? Preferably, we should only have $scope.isBatchImport. Also, I'm not too keen of using document.querySelector to check if it is a batch import. We should always try to avoid using the DOM functions in Angular.js.

src/main/resources/static/scss/components/_modals.scss
JorWo, line 1: I like the use of CSS animations here.

src/main/resources/static/javascript/mainApp/general.controller.js
yertsti: I think we need isLargeBatchImport cause it's used to trigger the warning model that you are trying to add more than 100 members either from the text box or a file. Maybe we could rename it to something else because batch import implies it came from an imported file. 
isBatchImport it is set in reader.onload and I had to add some clean up to set it to false depending on how addMembers() went. 

Also, I updated the large imports modal to look like the other add modals with the warning about how it takes 3-5 minutes for it to update.

JorWo:
If $scope.isLargeBatchImport controls displaying the import confirmation modal for imports with >100 members, what does $scope.isBatchImport control?

yertsti:
$scope.isBatchImport takes care of the error import modal. Without it, if there is an error while importing a file, like a file of 9 valid uuid and one invalid uuid, it would not add anyone and there would be no add-error-message warning box or no import error modal.

I could change it to where if there is an error with importing the users, it displays it with the add-error-message warning. If you are trying to add >100 users and there is an error, then we display the import error modal. That way we don't care about where the source of the members are coming from we just care about the amount of members being added.

JorWo:
> $scope.isBatchImport takes care of the error import modal. Without it, if there is an error while importing a file, like a file of 9 valid uuid and one invalid uuid, it would not add anyone and there would be no add-error-message warning box or no import error modal.
> 
> I could change it to where if there is an error with importing the users, it displays it with the add-error-message warning. If you are trying to add >100 users and there is an error, then we display the import error modal. That way we don't care about where the source of the members are coming from we just care about the amount of members being added.

I thought the idea you have already exists in our code. The only difference is the error is displayed based on the number of invalid uhIdentifiers thrown back. If there are 10 or less invalid uhIdentifiers then it is displayed with the add-error-message warning. Otherwise, display in the import error modal.

It seems like you can just have $scope.isBatchImport without $scope.isLargeBatchImport because as you mentioned, we do not care about where source of members are coming from and we only care about the number of members being added.

yertsti:
Alright, I'll get the import to display add-error-message on error.

JorWo:
One last thing, can you rename all $scope.isLargeBatchImport to $scope.isBatchImport? Thanks.

JorWo (approved):

PR #824, opened 2023-06-26
/admin - Manage Person row spacing correction

michho8:
# Ticket Link

[Ticket 1489](https://uhawaii.atlassian.net/browse/GROUPINGS-1489)

# List of squashed commits

- Get rid of `<p>` tags for smaller rows
- Change back to Bootstrap and change checkbox class
- Make sure checkbox inside borderline
- Change crown icon color to match external link icon

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:

### DIFF
--- a/src/main/resources/templates/admin.html
+++ b/src/main/resources/templates/admin.html
@@ -243,7 +243,7 @@
                                     <i class="fa fa-fw fa-external-link-alt clickable" tooltip="Manage grouping"
                                     ng-click="displayGroupingInNewTab(l.name, l.path)"></i>
                                     {{l.name}}
-                                    <i class="fa fa-crown fa-fw fa-pull-right clickable pt-1" role="button"
+                                    <i class="fa fa-crown fa-fw fa-pull-right clickable pt-1 text-primary" role="button"
                                        tooltip data-placement="right"
                                        tabindex="0"
                                        aria-label="Display admin."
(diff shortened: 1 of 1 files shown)


michho8:
Before:
image

After:
image

mhodgesatuh (approved):
Very clean; thanks!

JorWo:
Great work! I was wondering if the crown icons should be dark blue just like the color of the "open grouping in a new tab" button. My reasoning for that design is because all clickable buttons are in that color. One more thing to check is since you removed the `<p>` tags, is this manage person page still ADA compliant? Here is our documentation to test for ADA compliance: https://uhawaii.atlassian.net/wiki/spaces/SITARd/pages/12944279/American+Disabilities+Act+Compliance+Testing+Guide. The main tools you will need to use are Lighthouse and Axe DevTools.

michho8:
I can't seem to download the Chrome extension for Axe DevTools while on my school account because I don't have access to the Chrome Web Store.

For Lighthouse, the difference in score is in the "SEO" section where the test site has a score of 7/8 and the localhost site has a score of 6/7. Both have the same warning message: "Document does not have a meta description." In the "Performance" section, both sites have a score of 1/4 but the test site has 999 elements marked as "Avoid an excessive DOM size" whereas the localhost site has 987 elements. I'm not too sure if this is an improvement or not since the general score decreased but has fewer elements concerned, would like some feedback on this, thank you!

michho8:
Update: Axe DevTools produce no new errors.

JorWo:
> I can't seem to download the Chrome extension for Axe DevTools while on my school account because I don't have access to the Chrome Web Store.
> 
> For Lighthouse, the difference in score is in the "SEO" section where the test site has a score of 7/8 and the localhost site has a score of 6/7. Both have the same warning message: "Document does not have a meta description." In the "Performance" section, both sites have a score of 1/4 but the test site has 999 elements marked as "Avoid an excessive DOM size" whereas the localhost site has 987 elements. I'm not too sure if this is an improvement or not since the general score decreased but has fewer elements concerned, would like some feedback on this, thank you!

What matters is that the accessibility score is 100%. We can ignore the other scores.

JorWo (approved):
Thanks for making sure these changes are ADA compliant.

PR #826, opened 2023-06-29
Add icon to the nav bar - indicate user is not in their personal account

michho8:
# Ticket Link

[Ticket 1494](https://uhawaii.atlassian.net/browse/GROUPINGS-1494)

# List of squashed commits

- Create the icon and place it in the nav bar
- Create hover message for the whole icon
- Place icon only in department accounts
- Make "login" section inside the borderline
- Move icon next to UH Groupings logo
- Add string literal dynamicModal
- Split the logo into another section
- Get rid of message string literal
- LightHouse Accessibility and SEO score fixes
- Move icon to the right
- Add Department role
- Add in Department role into test file
- Address GitHub comments about testing

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/groupings/service/UhUuidCheckerService.java
+++ b/src/main/java/edu/hawaii/its/groupings/service/UhUuidCheckerService.java
@@ -1,12 +1,13 @@
 package edu.hawaii.its.groupings.service;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @Service
 public class UhUuidCheckerService {
@@ -22,5 +23,10 @@ public class UhUuidCheckerService {
 
         final Matcher matcher = pattern.matcher(uhUuid);
         return matcher.matches() || uhUuid.equals(uid);
+    }
+
+    public boolean isDepartmentAccount(String uhUuid, String uid) {
+        logger.info(String.format("isDepartmentAccount; uhUuid: %s; uid: %s", uhUuid, uid));
+        return uhUuid.equals(uid);
     }
 }
--- a/src/main/resources/templates/menubar.html
+++ b/src/main/resources/templates/menubar.html
@@ -1,13 +1,22 @@
 <div th:fragment="copy">
-<nav class="navbar fixed-top navbar-expand-lg navbar-light border-bottom pt-2 pb-2" th:fragment="copy">
+<nav class="navbar fixed-top navbar-expand-lg navbar-light border-bottom pt-2 pb-2" th:fragment="copy"
+     ng-controller="GeneralJsController">
     <div class="container">
         <a class="navbar-brand align-items-center" th:href="@{/}">
             <img class="img-fluid"
                  alt="UH Groupings Logo"
                  th:srcset="@{images/uh-groupings-logo.png 1x, images/uh-groupings-logo-2x.png 2x, images/uh-groupings-logo.svg}"
                  th:src="@{images/uh-groupings-logo.png}">
-            <span class="sr-only">UH Groupings</span>
-        </a>
+            <span class="sr-only">UH Groupings</span></a>
+        <span class="user-role rounded-circle seafoam-bg p-1 ml-5"
+              sec:authorize="hasAnyRole('ROLE_DEPARTMENT')"
+              tooltip data-placement="top"
+              th:title="#{screen.message.home.page.departmentAccount}">
+            <i class="fa fa-user p-2 position-relative" role="button" aria-label="Department Account Indicator"
+               th:attr="ng-click='displayDynamicModal(\'Warning\', \'' +
+                        #{screen.message.home.page.departmentAccount} + '\')'">
+                <i class="rounded-circle blue-bg text-light fa fa-school fa-xs p-1 position-absolute mt-2"></i>
+            </i></span>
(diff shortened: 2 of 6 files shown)


fduckart:
Looks good. Please change the hasAnyAuthority call to a hasAnyRole call, and then make sure it works. I realize there are authority calls in the file already, but I failed to notice that issue before.

michho8:
Ok, I changed the calls. There's one more `hasAnyAuthority` call, should I change that one as well? That's for the Groupings tab.

mhodgesatuh (commented):
Before approving this would like to see this presented at a team meeting.  Looking forward to seeing how it works.

JorWo:
I think the home page should have the departmental icon too:
image

michho8:
It was discussed at the team meeting that we should leave the two icons separate, thank you.

src/main/resources/templates/menubar.html
```diff
@@ -1,13 +1,22 @@
...
         <a class="navbar-brand align-items-center" th:href="@{/}">
             <img class="img-fluid"
                  alt="UH Groupings Logo"
                  th:srcset="@{images/uh-groupings-logo.png 1x, images/uh-groupings-logo-2x.png 2x, images/uh-groupings-logo.svg}"
                  th:src="@{images/uh-groupings-logo.png}">
-            <span class="sr-only">UH Groupings</span>
-        </a>
+            <span class="sr-only">UH Groupings</span></a>
+        <span class="user-role rounded-circle seafoam-bg p-1 ml-5"
+              sec:authorize="hasAnyRole('ROLE_ADMIN', 'ROLE_OWNER')"
```
JorWo, line 12: Is there anyway to make this specify specifically for a departmental account? Because right now, the icon is showing for me although I am logged into my personal account. I think it would helpful to have a test departmental account that we can log in with.

src/main/resources/templates/menubar.html
```diff
@@ -1,13 +1,22 @@
...
     <div class="container">
         <a class="navbar-brand align-items-center" th:href="@{/}">
             <img class="img-fluid"
                  alt="UH Groupings Logo"
                  th:srcset="@{images/uh-groupings-logo.png 1x, images/uh-groupings-logo-2x.png 2x, images/uh-groupings-logo.svg}"
                  th:src="@{images/uh-groupings-logo.png}">
-            <span class="sr-only">UH Groupings</span>
-        </a>
+            <span class="sr-only">UH Groupings</span></a>
+        <span class="user-role rounded-circle seafoam-bg p-1 ml-5"
```
JorWo, line 11: In my opinion, I think the departmental icon would look best flush against the UHGroupings logo, without ml-5 on large screens and mid-size screens:
image
However, the icon is currently floating in the middle with mid-size screens:
image
It should be flush on the left of the UHGroupings logo just like on large screens.

As for small screens, it looks fine:
image

mhodgesatuh:
Actually, I requested that it be centered in the white space between the logo on left and the nav links on the right since it has nothing to do with either, and I perceive that to make it a little more obvious.

michho8:
Created a Department role, waiting for feedback if this is a reasonable solution. I want to make this a role for future uses since we need to check if the `uhUuid == uid` which requires referencing the user and their information (I see a function `getUid` in UhAttributes.java). Would need to include that file in order to use the functions.

The builds and tests are failing because of AuthorizationServiceTest.java, just need to adjust the expected results in the tests because the Department role is true for some of the test cases. Should be a quick adjustment and will finish if creating a department role is an acceptable solution.

mhodgesatuh (approved):

src/main/java/edu/hawaii/its/groupings/service/UhUuidCheckerService.java
```diff
@@ -23,4 +24,9 @@ public boolean isValidUhUuid(String uhUuid, String uid) {
         final Matcher matcher = pattern.matcher(uhUuid);
         return matcher.matches() || uhUuid.equals(uid);
     }
+
+    public boolean isDepartmentAccount(String uhUuid, String uid) {
+        logger.info("isDepartmentAccount; returns if uhUuid match uid");
```
JorWo, line 29: For this logger message as well as in the isValidUhUuid() function, could you have the message show the uhUuid and uid passed in? This piece of code should work: 
`logger.info(String.format("isDepartmentAccount; uhUuid: %s; uid: %s;, uhUuid, uid));`
Thanks!

src/test/java/edu/hawaii/its/groupings/access/AuthorizationServiceTest.java
```diff
@@ -99,6 +99,7 @@ public void fetchPrivilegesTest() {
         assertThat(roleHolder.size(), is(4));
         assertTrue(roleHolder.contains(Role.ANONYMOUS));
         assertTrue(roleHolder.contains(Role.UH));
+        assertFalse(roleHolder.contains(Role.DEPARTMENT));
```
JorWo, line 102: In the lines above, there are mocks that fake a function to return true. Could you mock the isDepartmentAccount() function to return true then assertTrue that roleHolder.contains(Role.DEPARTMENT). This is just to match the rest of the tests to assertTrue. Thanks.

michho8: I reassigned the value for uhUuid in order to use assertTrue in this part of the test.
```
uhUuid = "test";
roleHolder = authorizationService.fetchRoles(uhUuid, "test");
assertTrue(roleHolder.contains(Role.DEPARTMENT));

src/main/java/edu/hawaii/its/groupings/service/UhUuidCheckerService.java
michho8: The code provided works, thank you.

JorWo (approved):
Great work!

PR #837, opened 2023-07-19
Updating Export Grouping Button

Kristyn-Mimura:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1448)

# List of squashed commits

- Add new option to export grouping 
- Makes button looks more like dropdown
- Makes isGroupingEmpty function

# Test Checklist

- [X] Unit Tests Passed:
- [X] Jasmine Tests Passed:


mhodgesatuh (approved):
Thanks!

Wasn't expecting the following addition, but assuming that you found another fix while you were at it:
   import edu.hawaii.its.groupings.exceptions.PasswordFoundException;

src/main/resources/templates/fragments/selected-grouping.html
```diff
@@ -10,41 +10,43 @@
...
-                    Export Members
-                </button>
-            </div>
-            <button type="button" class="btn btn-primary dropdown-toggle-split dropdown-toggle" data-toggle="dropdown"
-                    aria-label="More export options" aria-haspopup="true" aria-expanded="false"
-                    ng-disabled="!paginatingComplete || (groupingMembers.length === 0 && groupingExclude.length === 0)">
+        <div class="btn-group dropdown" id="csvButton" tooltip data-original-title="{{ getCSVToolTipMessage() }}">
+            <button type="button" class="form-control" data-toggle="dropdown"
+                    aria-haspopup="true" aria-expanded="false"
+                    ng-disabled="!paginatingComplete || (groupingMembers.length === 0 && groupingBasis.length === 0 && groupingInclude.length === 0 && groupingExclude.length === 0)">
```
JorWo, line 18: Could you put this long condition: `(groupingMembers.length === 0 && groupingBasis.length === 0 && groupingInclude.length === 0 && groupingExclude.length === 0)` into its own function? I think isGroupingEmpty() would be a good name. Also add a Jasmine test for it. Thanks!

src/main/resources/templates/fragments/selected-grouping.html
```diff
@@ -10,41 +10,43 @@
...
+                <i class="fa fa-file-export mr-1"></i>
+                Export Grouping
```
JorWo, line 20: I like the new design of the dropdown button but I also think adding a caret/chevron would make the design even better because it tells the user that this is a dropdown. Here are some designs with a caret or a chevron:

Caret:
image
I used this code after the "Export Grouping" text: `<i class="fa fa-caret-down ml-1"></i>`

Chevron:
image
I used this code after the "Export Grouping" text: `<i class="fa fa-chevron-down ml-1"></i>`

Let me know what you think.

mhodgesatuh:
Personally I am 51/49, leaning just barely toward the smaller caret merely
for the sake of its compactness.
I also think I see the caret more commonly.

src/main/resources/templates/fragments/selected-grouping.html
Kristyn-Mimura: I lean more towards the caret, the chevron is used on the site for toggling the alphabetical order of names in the tables so using the caret would differentiate between the two functionalities. Though I personally find the chevron more aesthetically pleasing

JorWo: Ok let's go with the caret.

src/test/javascript/grouping.controller.test.js
```diff
@@ -2710,4 +2710,19 @@ describe("GroupingController", () => {
...
+            expect(scope.isGroupingEmpty()).toEqual(false);
+        });
+
+        it("should return true if there are no members in a grouping", () => {
+            scope.groupingMembers = [];
+            scope.groupingInclude = [];
+            scope.groupingExclude = [];
+            scope.groupingBasis = [];
+
+            expect(scope.isGroupingEmpty()).toEqual(true);
```
JorWo, line 2725: You can use .toBeTrue() and .toBeFalse() instead of .toEqual() here.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -1753,6 +1753,13 @@
             }
             return str;
         };
+
+        /**
+         * Check if there are no members in the grouping
+         */
+        $scope.isGroupingEmpty = () => {
+            return ($scope.groupingMembers.length === 0 && $scope.groupingBasis.length === 0 && $scope.groupingInclude.length === 0 && $scope.groupingExclude.length === 0);
+        };
```
JorWo, line 1762: This line is bit long, could you add some newlines at the front of each `&&`? Also the parenthesis can be removed too. Thanks!

JorWo (approved):
Looks great!

PR #870, opened 2023-11-12
Fix using trashcan to remove old test accounts from Admins

imths:
# Ticket Link

[Groupings-1568](https://uhawaii.atlassian.net/browse/GROUPINGS-1568?atlOrigin=eyJpIjoiNjVmZTQ1NDk5MWU5NGI2YTg2NmExNzVjMWExNmZiNzIiLCJwIjoiaiJ9)

# List of squashed commits

- None

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -942,7 +942,7 @@
             if (/[0-9]{8}/.test(memberIdentifier)) {
                 memberToReturn = _.find(currentPage, (member) => member.uhUuid === memberIdentifier);
             } else {
-                memberToReturn = _.find(currentPage, (member) => member.uid === memberIdentifier);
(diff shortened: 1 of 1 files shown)


mhodgesatuh (commented):
Looks like another nice catch.  I'll need to also see this in action before approving though.

JorWo:
This change is a bit weird, it makes it so that the last else statement can never be reached.

mhodgesatuh (commented):
the member.username reference is a red flag for me since we've agreed to replace the use of username for variables and attributes to uid.  Does the member object need to be updated?  That's how it appears to me at first glance.

JorWo:
I believe our API response for retrieving the list of Admins still uses
username as the uid. Misha is currently working on ticket to update that.

mhodgesatuh:
Sounds good.  We can push this PR through and wait for the next to make the
correction.
Thanks

JorWo (approved):
Looks good!

PR #879, opened 2023-12-03
Display planned outage announcement on the home page

michho8:
# Ticket Link

[Ticket 974](https://uhawaii.atlassian.net/browse/GROUPINGS-974)

# List of squashed commits

- Change endpoint name
- Console logs and getting rid of head message
- Get rid of white spacing on top
- Message at top
- Finalize position of alert message
- Rename endpoint and functions
- Fix home page display
- Commented out files
- Display planned outage announcement on the home page
- Messages are aligned and within page dimensions
- Clean up code and add API error modal
- Add getOutageMessages test
- Fix test for announcements path
- Separate alert messages and add close functionality
- Dismissable alerts and test for controller
- Get rid of white space between alerts
- Add back call to the endpoint
- Merge conflicts with main
- Rebranch
- Fix home page format
# Test Checklist

- [ ] Project Jasmine Tests Passed:


src/main/java/edu/hawaii/its/groupings/configuration/SecurityConfig.java
```diff
@@ -184,6 +182,7 @@ public SecurityFilterChain filterChain(HttpSecurity http, CasAuthenticationFilte
 
         http.authorizeHttpRequests((auths) -> auths
                 .requestMatchers(antMatcher("/")).permitAll()
+                .requestMatchers(antMatcher("/announcements/active")).permitAll()
```
JorWo, line 185: We can go with setting the `antMatcher` to "/annoucements/** so that it opens more possibilities in the future to filter the announcements by other statuses.

src/main/java/edu/hawaii/its/groupings/controller/OutageRestController.java
JorWo, line 1: Let's call this controller the AnnoucnementsController and make sure the import order has been optimized in the class. Also remove the newline after `@RestController` on line 16 and add a newline after `private HttpRequestService httpRequestService;`. Thanks!

JorWo: Actually, is having a separate controller necessary? If not, let's keep it consistent between the UI and API to have the the announcements endpoints stored in the GroupingsRestController.

mhodgesatuh: I started a thread in slack.  We need to have a basic architectural question from which we derive a guiding principle.

JorWo: Sorry, I just remembered this controller must be separate so that it can be called without needing to be logged-in. I think it would be a good idea to make this the new controller you created with Frank from last Thursday's meeting.

src/main/java/edu/hawaii/its/groupings/configuration/SecurityConfig.java
michho8: The /active is to indicate the status of the messages. In a previous discussion, it was decided that we want to create other endpoints with "/future" or "/expired."

src/main/java/edu/hawaii/its/groupings/controller/OutageRestController.java
michho8: Ok, I'll rename this controller to be generalized. How about `GeneralRestController` ? This will be for all endpoints that should be available to the general audience, i.e. don't need to be logged in but will still work if logged in.

JorWo: I would name it `OpenRestController` so that it is not confused to be with the general.controller.js in the JS.

JorWo: I also think it would be very nice if you could find a way put the announcements endpoint in the GroupingsRestController and by messing with the requestMatchers in the SecurityConfig. That way both UI and API projects have the same controllers.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
```diff
@@ -644,6 +646,19 @@ public ResponseEntity<String> getAsyncJobResult(Principal principal, @PathVariab
...
 
+    /**
+     * Get the list of active announcements to display.
+     */
+    @GetMapping(value = "/announcements/active")
+    public ResponseEntity<String> getAnnouncements() {
+        if (!API_HANDSHAKE_ENABLED) {
+            logger.info("API Handshake is not enabled.");
+            return new ResponseEntity<>(JsonUtil.asJson(""), HttpStatus.OK);
+        }
```
JorWo, line 657: Is this if statement needed? Also add the logger.info("Entered REST getAnnouncements...") to this.

src/main/java/edu/hawaii/its/groupings/configuration/SecurityConfig.java
JorWo: Perfect, could you set it to "/accouncements/**" now for the convenience of the person who will implement the other endpoints.

JorWo:
Also change the functions named `getAnnouncements` to `activeAnnouncements` to match the API function names, thanks.

src/main/java/edu/hawaii/its/groupings/controller/OutageRestController.java
michho8: I put back the separate controller because this was causing the `MissingResourceException`. I believe that is because the endpoint was looking just for "announcements/active" but the GroupingsRestController starts off with "api/groupings/". I tried to adjust the endpoint call to include `BASE_URL` in front but that still gave another error related to the CORS issue with the currentUser call.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
michho8: Took out those lines and added the log, thank you.

JorWo (approved):
Looks great!

mhodgesatuh (approved):
Looks good!  thanks

PR #895, opened 2024-01-19
Rewrite checkMainSelectAllCheckbox with more clarity

michho8:
# Ticket Link

[Ticket 1611](https://uhawaii.atlassian.net/browse/GROUPINGS-1611)

# List of squashed commits

- Simplify the conditional
- Add comments to the for loop
- Address pr suggestions

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -1668,7 +1668,7 @@
                 }
                 let check = $scope.checkMainSelectAllCheckbox(item);
                 $scope.pageSelected = check.page;
-                $scope.showToolIcon = check.toolTip;
+                $scope.showToolIcon = check.tooltip;
             } else {
                 if ($scope.paginationPageChange === true) {
                     $scope.updateCheckboxEventListeners(item, group);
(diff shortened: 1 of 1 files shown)


JorWo (approved):
Looks good, could you also make the t in check.toolTip lowercase? Thanks.

mhodgesatuh (approved):
trueFalseCount confused me for a moment, but makes sense now.  itemsCheckedCount.checked and itemsCheckedCount.notChecked seems to tie in better with the item.allItems reference.  Another idea is to overload item such that you have item.totalChecked and item.totalNotCheck.  I'll approve it, but I think the use of "item" or "items" could be better normalized here.

michho8:
I assume the suggestion for `itemsCheckedCount.checked` and `itemsCheckedCount.notChecked` is for my usage for `trueFalseCount.true`. With the `_.countBy` function, it returns two properties: "true" and "false" which returns an integer representing how many items have been checked and not checked. I changed it by making `itemsCheckedCount` an object with `checked` and `notChecked` as properties set to the `_.countBy` function.

I decided to change this portion and not to overload the `item` parameter as that would mean having to deal with an outside function since it is initialized there. Thank you for the suggestions.

PR #900, opened 2024-01-25
Write more Jasmine tests for dynamic select-all checkboxes feature

michho8:
# Ticket Link

[Ticket 1599](https://uhawaii.atlassian.net/browse/GROUPINGS-1599)

# List of squashed commits

- Write tests for toggleCheckAllSelection
- Starting checkMainSelectAllCheckbox tests
- Testing if Select All checkbox is unchecked
- Testing with more than 20 users checked
- Align Select All checkbox with others
- Change let to const

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -1647,15 +1647,14 @@
          *                                   If "page", it toggles checkboxes on the current page; if "other", it updates event listeners.
          */
         $scope.toggleCheckAllSelection = (group, checkboxClicked) => {
-            let item;
             if (group === "Exclude") {
-                item = {
+                $scope.item = {
                     allItems: $scope.groupingExclude,
                     pageItems: $scope.pagedItemsExclude,
                     pageNumber: $scope.currentPageExclude
                 };
             } else if (group === "Include") {
-                item = {
+                $scope.item = {
                     allItems: $scope.groupingInclude,
                     pageItems: $scope.pagedItemsInclude,
                     pageNumber: $scope.currentPageInclude
@@ -1663,15 +1662,15 @@
             }
             if (checkboxClicked === "page") {
                 $scope.pageSelected = !$scope.pageSelected;
-                for (let member of item.pageItems[Number(item.pageNumber)]) {
+                for (const member of $scope.item.pageItems[$scope.item.pageNumber]) {
                     $scope.membersInCheckboxList[member.uhUuid] = $scope.pageSelected;
                 }
-                let check = $scope.checkMainSelectAllCheckbox(item);
+                const check = $scope.checkMainSelectAllCheckbox($scope.item);
                 $scope.pageSelected = check.page;
                 $scope.showToolIcon = check.tooltip;
             } else {
                 if ($scope.paginationPageChange === true) {
-                    $scope.updateCheckboxEventListeners(item, group);
+                    $scope.updateCheckboxEventListeners($scope.item, group);
                     $scope.paginationPageChange = false;
                 }
             }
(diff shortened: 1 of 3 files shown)


JorWo:
I think every `let` used in this PR can be replaced with `const`.

michho8:
Just changed `let` to `const` in my tests, thank you.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -1647,31 +1647,30 @@
...
                 };
             }
             if (checkboxClicked === "page") {
                 $scope.pageSelected = !$scope.pageSelected;
-                for (let member of item.pageItems[Number(item.pageNumber)]) {
+                for (let member of $scope.item.pageItems[Number($scope.item.pageNumber)]) {
                     $scope.membersInCheckboxList[member.uhUuid] = $scope.pageSelected;
                 }
-                let check = $scope.checkMainSelectAllCheckbox(item);
+                let check = $scope.checkMainSelectAllCheckbox($scope.item);
```
JorWo, line 1668: These two lines can be const too.

src/test/javascript/grouping.controller.test.js
```diff
@@ -2529,41 +2529,122 @@ describe("GroupingController", () => {
...
+                    name: `User ${i}`,
+                    uid: `user${i}`,
+                    uhUuid: `0000000${i}`,
+                    firstName: "User",
+                    lastName: `${i}`
+                };
+                scope.testItem.pageItems[0].push(newItem);
+            }
+            // Set all users to checked
+            for (let item of scope.testItem.pageItems[0]) {
```
JorWo, line 2640: And the last change to const is here.

src/main/resources/static/javascript/mainApp/grouping.controller.js
michho8: I changed line 1668 and got rid of that Number type converter. Is there another place I'm missing? You mentioned that there are two lines to change `let` to `const`.

michho8: Oh, now that I saw your other comment, you're referring to inside the `for` loop. I didn't know you could use `const` in this syntax for iteration, thank you.

JorWo (approved):
Everything looks good, thanks for writing the Jasmine tests.

mhodgesatuh (approved):
Given how tricky this code implementation seemed to be in the past, very glad to see these tests.  Thanks.

Is there a test to see that the All checkbox is unchecked if one of the below checkboxes are unchecked?

JorWo:
> Given how tricky this code implementation seemed to be in the past, very glad to see these tests. Thanks.
> 
> Is there a test to see that the All checkbox is unchecked if one of the below checkboxes are unchecked?

That scenario would be great to test in Selenium.

src/test/javascript/grouping.controller.test.js
```diff
@@ -2529,41 +2529,122 @@ describe("GroupingController", () => {
...
+        });
+
+        it("should return the original check values", () => {
+            scope.testItem.pageItems = [];
+            const testCheck = scope.checkMainSelectAllCheckbox(scope.testItem);
+            expect(testCheck.page).toBeTrue();
+            expect(testCheck.tooltip).toBeFalse();
+        });
+
+        it("should uncheck the Select All checkbox if there are any unchecked checkboxes", () => {
```
michho8, line 2614: There is this test that manually writes one false value in the `scope.membersInCheckboxList` which is the equivalent of unchecking a box. I'm not familiar with Selenium testing but think that should be a different ticket since this is for our Jasmine tests.

PR #907, opened 2024-02-13
Fix the Select Page checkbox bug and rewrite

michho8:
# Ticket Link

[Ticket 1613](https://uhawaii.atlassian.net/browse/GROUPINGS-1613)

# List of squashed commits

- Reflect changes onto exclude.html
- Remove counting members to determine if tooltip shown
- Changed checkMainSelectAllCheckbox Jasmine test
- Add checkSingleBox Jasmine tests
- Fix tooltip appearance
- Enter key works for Select All and single checkboxes
- Change function names and reflect changes to exclude.html
- Fix tooltip bug
- Get rid of unnecessary checkSelectAll function call
- Create constant tooltip variable
- Fix the Enter key bug and create an additional function to store duplicate code 
- Jasmine tests for Select All and single toggle checkbox
- WIP assignListToGroup Jasmine test
- Finish Jasmine tests 
- Address two pr requests
- Rename the item object in controller
- Rename the item object in Jasmine tests
- Address Codacy complaints
- Fix Object Injection Sink Codacy complaint

# Test Checklist


JorWo (approved):
Looks much better!

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -1744,46 +1755,29 @@
         };
 
         /**
-         * Checks the state of the main "Select All" checkboxes and tooltip based on checkbox changes.
+         * Checks the state of the main "Select All" checkbox based on checkbox toggles.
          *
          * @param {Object} item - An object containing information about the group and pagination.
```
michho8, line 1760: I would like to change the name of the `item` object to reflect a more accurate description. I'm thinking of `groupMembers` because the three properties all revolve around the members within the group.

1. `allItems` - all of the members within a group
2. `pageItems` - all of the members within the page (max of 20 per page)
3. `pageNumber` - the page number that the user currently is on (but we can take this out because it isn't being used anymore since we assume the user will see the tooltip and remove all of the members at once)

What do you think about the name change and getting rid of the `pageNumber` property?

JorWo:
Can you do two favors for me?
1. Put mainv2_3.css.map in the `.gitignore`, then delete that file from this PR. The mainv2_3.css gets autogenerated every time the Spring Boot app runs with our new SCSS compiler. 
2. Delete the console.log() inside `announcement.controller.js`
Thanks!

src/main/resources/static/javascript/mainApp/grouping.controller.js
JorWo: That sounds good.

JorWo: I think it was named item in the first place because we have a $scope variable named $scope.item, but this function does not use that $scope variable.

mhodgesatuh (approved):

src/main/resources/static/javascript/mainApp/grouping.controller.js
michho8: Decided to leave the `pageNumber` property as it was being used in the event listener functions as well.

PR #936, opened 2024-05-07
Implement default user login without CAS authentication

gitCarrot:
# Ticket Link

[Ticket-1670](https://uhawaii.atlassian.net/browse/GROUPINGS-1670)

# List of squashed commits

- Create ootb security config with custom filter chain

- Create authentication filter for bypass authentication that extends GerernicFilterBean and OotbCustomUserDetails for having ootb default suer information

- Add the logic to make OotbStaticUserAuthenticationFilter bean with the ootb default user properties from application-ootb and pass it through SecurityFilterChain

- Add ConditionalOnProperty grouping.api.server.type=GROUPER for security config

- Add and modify the ootb default user properties in application-ootb.properties, and manage the two types of API servers separately using two properties files

- Create the ootb security config test
- Fix codacy code style and environment of OotbSecurityConfigTest to skip the cas login related properties
- Add private to the field for ootb user properties and requestMatchers same as existed security config
- Fix codes with reviews
- Put all the ootb.default properties in a single class and modified all the related constructions
- Implement OotbCustomUserDetailsTest and OotbStaticUserAuthenticationFilterTest for checking security context and construction of OotbCustomUserDetails object
- Apply uh groupings code style
- Organize imports
- Fix mis typo

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/groupings/configuration/OotbCustomUserDetails.java
+++ b/src/main/java/edu/hawaii/its/groupings/configuration/OotbCustomUserDetails.java
@@ -1,0 +1,25 @@
+package edu.hawaii.its.groupings.configuration;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.security.core.GrantedAuthority;
+
+import edu.hawaii.its.groupings.access.UhAttributes;
+import edu.hawaii.its.groupings.access.User;
+
+public class OotbCustomUserDetails extends User {
+
+    public OotbCustomUserDetails(String username, String password, Collection<GrantedAuthority> authorities,
+            String givenName, String cn, String email) {
+        super(username, password, authorities);
+        Map<String, Object> defaultAttributes = new HashMap<>();
+        defaultAttributes.put("cn", Arrays.asList(cn));
+        defaultAttributes.put("mail", Arrays.asList(email));
+        defaultAttributes.put("givenName", Arrays.asList(givenName));
+        this.setAttributes(new UhAttributes(defaultAttributes));
+    }
+
+}
--- a/src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
+++ b/src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
@@ -1,0 +1,110 @@
+package edu.hawaii.its.groupings.configuration;
+
+import static org.springframework.security.config.Customizer.withDefaults;
+import static org.springframework.security.web.util.matcher.AntPathRequestMatcher.antMatcher;
+
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.web.SecurityFilterChain;
+import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
+import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
+import org.springframework.ws.config.annotation.DelegatingWsConfiguration;
+
+import edu.hawaii.its.groupings.access.DelegatingAuthenticationFailureHandler;
+
+@EnableWebSecurity
+@Configuration
+@ConditionalOnProperty(name = "grouping.api.server.type", havingValue = "OOTB")
+public class OotbSecurityConfig {
+
+    private static final Log logger = LogFactory.getLog(OotbSecurityConfig.class);
+
+    @Value("${grouping.api.server.type}")
+    private String serverType;
+
+    @Value("${ootb.default.user.name}")
+    private String name;
+
+    @Value("${ootb.default.user.username}")
+    private String username;
+
+    @Value("${ootb.default.user.authorities}")
+    private List<String> authorities;
+
+    @Value("${ootb.default.user.cn}")
+    private String cn;
+
+    @Value("${ootb.default.user.email}")
+    private String email;
+
+    @Value("${url.base}")
+    private String appUrlBase;
+
+    @Bean
+    public AuthenticationManager authenticationManager(AuthenticationConfiguration authenticationConfiguration)
+            throws Exception {
+        return authenticationConfiguration.getAuthenticationManager();
+    }
+
+    @Bean
+    public DelegatingAuthenticationFailureHandler authenticationFailureHandler() {
+        return new DelegatingAuthenticationFailureHandler(appUrlBase);
+    }
+
+    @Bean
+    public DelegatingWsConfiguration delegatingWsConfiguration() {
+        return new DelegatingWsConfiguration();
+    }
+
+    @Bean
+    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        http
+                .addFilterBefore(ootbStaticUserAuthenticationFilter(), UsernamePasswordAuthenticationFilter.class)
+                .authorizeHttpRequests((auths)
+                        -> auths
+                        .requestMatchers(antMatcher("/")).permitAll()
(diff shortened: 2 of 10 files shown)


mhodgesatuh (approved):

src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
```diff
@@ -0,0 +1,80 @@
...
+    String serverType;
+
+    @Value("${ootb.default.user.name}")
+    String name;
+
+    @Value("${ootb.default.user.username}")
+    String username;
+
+    @Value("${ootb.default.user.authorities}")
+    List<String> authorities;
```
JorWo, line 41: Make sure to add the `private` keyword before the variables. Also there should be a newline after line 41, `List<String> authorities;`. Thanks!

src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
```diff
@@ -0,0 +1,80 @@
...
+        return new DelegatingWsConfiguration();
+    }
+
+    @Bean
+    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        http
+                .addFilterBefore(ootbStaticUserAuthenticationFilter(), UsernamePasswordAuthenticationFilter.class)
+                .authorizeHttpRequests((auths)
+                        -> auths.requestMatchers(antMatcher("/")).permitAll()
+                        .anyRequest().authenticated())
```
JorWo, line 67: Question, is there a reason that you chose all request matchers to be permit all?

gitCarrot: I wrote that code in the very beginning to make it easy to implement bypass authentication before I was able to set default user authorities. Since we can set the default user authorities through properties file now, I think I can make it all request matchers same as current existed security config. Thanks!

gitCarrot: Everything works well in the ootb ui after adding same requestMatchers as security config class.

src/main/java/edu/hawaii/its/groupings/configuration/OotbStaticUserAuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        this.serverType = serverType;
+        this.name = name;
+        this.username = username;
+        this.authorities = authorities;
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain)
+            throws IOException, ServletException {
+        if ("OOTB".equals(serverType) && SecurityContextHolder.getContext().getAuthentication() == null) {
```
JorWo, line 33: Could you invert the the `.equals` from `"OOTB".equals(serverType)` to `serverType.equals("OOTB")`? Thanks!

src/main/resources/application-ootb.properties
```diff
@@ -0,0 +1,17 @@
+# =========================================================================
+# Spring-related.
+spring.config.activate.on-profile=ootb
+spring.config.import=optional:file:${user.home}/.${user.name}-conf/uh-groupings-ui-overrides.properties
+app.api.handshake.enabled=false
+
+# There are two options: GROUPER, OOTB : "GROUPER" is a real grouper api service while "OOTB" is a ootb grouper api service.
+# You can change active profiles from the RUN/DEBUG Configurations.
```
JorWo, line 8: The comment here is not very clear to me about changing active profiles. I'm assuming this is about adding the flag in `./mvnw clean spring-boot:run` to the environment to be dev, test, prod. etc. Maybe this can be removed since the API side does not have this comment.

src/main/resources/application.properties
```diff
@@ -23,6 +23,9 @@ cas.send.renew=false
 # =========================================================================
 # Grouper related.
 app.groupings.controller.uuid=1d7365a23c994f5f83f7b541d4a5fa5e
+# There are two options: GROUPER, OOTB : "GROUPER" is a real grouper api service while "OOTB" is a ootb grouper api service.
+# You can change active profiles from the RUN/DEBUG Configurations.
```
JorWo, line 27: Same as the comment I left in application-ootb.properties.

src/main/java/edu/hawaii/its/groupings/configuration/OotbCustomUserDetails.java
```diff
@@ -0,0 +1,24 @@
...
+import edu.hawaii.its.groupings.access.UhAttributes;
+import edu.hawaii.its.groupings.access.User;
+
+public class OotbCustomUserDetails extends User {
+    public OotbCustomUserDetails(String username, String password, Collection<GrantedAuthority> authorities,
+            String givenName) {
+        super(username, password, authorities);
+        Map<String, Object> defaultAttributes = new HashMap<>();
+        defaultAttributes.put("cn", Arrays.asList("James"));
+        defaultAttributes.put("mail", Arrays.asList("example@example.com"));
```
JorWo, line 19: These two attrributes should be set in the properties file too. I'm not sure if the "mail" is being used by the Feedback page. Would be interesting to know.

fduckart:
All the diffs because of formatting... it's kind of a bummer.

mhodgesatuh:
and avoidable

fduckart:
So sorry, I've been out of the loop, but why are we doing this again?

gitCarrot:
> All the diffs because of formatting... it's kind of a bummer.

Just realized formatting for ui grouping project is kind of broken. I think I need to reimport the file for formatting

gitCarrot:
This ticket is for bypass authentication for ootb users.

src/main/resources/application-ootb.properties
gitCarrot: I'm going to add this feature in the api project default user related ticket too.

You can change the propfile like this below.

gitCarrot: and then if you do `./mvnw clean spring-boot:run` it's going to use active file that we set

src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
```diff
@@ -0,0 +1,102 @@
...
+    private String serverType;
+
+    @Value("${ootb.default.user.name}")
+    private String name;
+
+    @Value("${ootb.default.user.username}")
+    private String username;
+
+    @Value("${ootb.default.user.authorities}")
+    private List<String> authorities;
```
JorWo, line 40: I think you should put all these variables into the OotbCustomUserDetails class. Then the variables are all in one place rather than being split between two classes. Then at the `ootbStaticUserAuthenticationFilter()` bean, It can simply be: `return new OotbStaticUserAuthenticationFilter();`

JorWo, line 40: Could you find a way to put all the ootb.default properties all in one rather class rather than being split into two classes? Thanks!

JorWo:
Make sure to write test classes for OotbCustomUserDetails.java and OotbStaticUserAuthentication.java. They can be simple tests, like using the constructors and checking that the object created is not null. The doFilter() method in OotbStaticUserAuthenticationFilter can be ignored if you find that it is too difficult to unit test.

src/main/java/edu/hawaii/its/groupings/configuration/OotbSecurityConfig.java
gitCarrot: Nice suggestion thank you

src/test/java/edu/hawaii/its/groupings/configuration/OotbStaticUserAuthenticationFilterTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+import java.io.IOException;
+import java.util.List;
+
+import static org.mockito.Mockito.*;
```
JorWo, line 18: Make sure to remove * imports from all files. Also it seems that your code style is not enabled because static imports are usually on the top. Thanks!

JorWo: Replace the all import here from Mockito with the imports for when(), verify(), thenReturn(), never(), and any().

gitCarrot: Oh I don't know why it was not updated by git push, but I fixed it

JorWo (approved):
Looks great!

PR #941, opened 2024-06-13
Fix 'Manage My Groupings Fleeting Pagination Presentation Bug'

mikhail-shkaralevich:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1703)

# List of squashed commits

- Fix 'Manage My Groupings Fleeting Pagination Presentation Bug'

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -166,6 +166,7 @@
             const groupingPath = $scope.selectedGrouping.path;
             const paths = [groupingPath + ":basis", groupingPath + ":include", groupingPath + ":exclude", groupingPath + ":owners"];
             let currentPage = 1;
+            $scope.setPage('First', 'currentPageMembers', 'pagedItemsMembers');
             $scope.loading = true;
             $scope.paginatingComplete = false;
             await $scope.getGroupingDescription(groupingPath);


mikhail-shkaralevich:
Please note that I fixed the issue by setting the page to the first one when switching back to the "List all members" window. Is this the behavior we would expect?

mikhail-shkaralevich:
Also, I have a question, what's the point of calling `reloadGrouping()` when we switch to the "List all members" window from another window? It just seems to reload what's already loaded.

JorWo:
> Also, I have a question, what's the point of calling `reloadGrouping()` when we switch to the "List all members" window from another window? It just seems to reload what's already loaded.

We wanted the grouping to always stay updated by making it double as the reload button, but I have noticed some side-effects where I did want to reload the grouping because the grouping was large. I would really like to have an individual reload button instead.

JorWo (approved):
That's really smart and intuitive, great work.

JorWo:
> Please note that I fixed the issue by setting the page to the first one when switching back to the "List all members" window. Is this the behavior we would expect?

Yeah that seems to make a lot of sense; reloading the grouping should reset you to the first page.

mhodgesatuh:
That's an interesting idea.  If you write up a ticket we can add that
feature n 2.6 or 2.7

JorWo:
One question though, did you find any side-effects with this approach? Maybe you cannot change pages when a large grouping is loading?

mhodgesatuh (approved):
Impressive that the change is this light.

mhodgesatuh:
Note that the ticket is actually for 2.5

mikhail-shkaralevich:
> Note that the ticket is actually for 2.5

Yes, I noticed it but the bug was still present.

mikhail-shkaralevich:
> One question though, did you find any side-effects with this approach? Maybe you cannot change pages when a large grouping is loading?

I still can change pages while it is loading.

mikhail-shkaralevich:
> That's an interesting idea. If you write up a ticket we can add that feature n 2.6 or 2.7





> > Also, I have a question, what's the point of calling `reloadGrouping()` when we switch to the "List all members" window from another window? It just seems to reload what's already loaded.
> 
> We wanted the grouping to always stay updated by making it double as the reload button. Sometimes there could be two people making changes to the grouping at once, so we will need to reload the grouping. However I have noticed some side-effects where I did not want to reload the grouping because the grouping was large. I would really like to have an individual reload button instead.

Would the button be accessible from all windows of the "manage grouping page"?

mhodgesatuh:
I've not thought about the UI/UX.  So someone needs to prose it. Maybe good
discussion for the next meeting?

mikhail-shkaralevich:
> That's an interesting idea. If you write up a ticket we can add that feature n 2.6 or 2.7

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1709)

mhodgesatuh:
Thanks, I updated the ticket, and also moved it out of the parking lot.  If
you are working on it you still need to move it to the "in progress" swim
lane.

JorWo:
Make sure to backport to devrc-2-5-0.

PR #945, opened 2024-06-15
Add a 'Refresh' button in the 'Manage grouping' window

mikhail-shkaralevich:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1709)

# List of squashed commits

- Add a 'Refresh' button in the 'Manage grouping' window

# Test Checklist

### DIFF
--- a/src/main/resources/templates/fragments/members.html
+++ b/src/main/resources/templates/fragments/members.html
@@ -11,6 +11,14 @@
                      role="status">
                     <span class="sr-only">Loading...</span>
                 </div>
+                <i class="refresh-button fas fa-sync-alt ml-1 pr-2 d-inline-block"
+                   ng-show="!paginatingProgress && !loading"
+                   ng-click="reloadGrouping();resetFields()"
+                   role="button"
+                   aria-label="Refresh grouping"
+                   tooltip data-placement="right"
+                   th:title="#{screen.message.common.refreshMembers}">
+                </i>
                 <p class=" mb-1" ng-show="largeGrouping" th:text="#{screen.message.common.loading.toolarge}"><i
                         class="fa fa-check" aria-hidden="true"></i></p>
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/fragments/members.html
```diff
@@ -11,6 +11,14 @@ <h1 class="font-weight-bold text-dark pt-2 mb-0 d-inline-block">All Members
                      role="status">
                     <span class="sr-only">Loading...</span>
                 </div>
+                <i class="fas fa-sync-alt fa-lg ml-1 pr-2 d-inline-block"
+                   ng-show="!paginatingProgress && !loading"
+                   ng-click="reloadGrouping();resetFields()"
+                   role="button"
+                   aria-hidden="true"
```
JorWo, line 18: It would be better to use an ARIA label like `aria-label="Refresh grouping"` instead of `aria-hidden`. ARIA attributes give context to screen readers so that they know exactly what an element does. This is needed especially for this case when a button does not have text.

JorWo:
I think it would look great if the reload icon matches the size of the loading spinner: 
image
image

mhodgesatuh (approved):

src/main/resources/templates/fragments/members.html
mikhail-shkaralevich: Ok, thank you for letting me know!

mikhail-shkaralevich:
> I think it would look great if the reload icon matches the size of the loading spinner

Should I make it spin on hower?

JorWo:
That definitely sounds like a cool idea but at the same time might be
confusing to the user because they might mistake the hover animation as
actually reloading the grouping.

JorWo (approved):
Looks good!

PR #946, opened 2024-06-19
Refactoring the logic related to the button to update active user profiles in ootb

gitCarrot:
# Ticket Link

[Ticket-1712](https://uhawaii.atlassian.net/browse/GROUPINGS-1712)

# List of squashed commits

- Implement loadAvailableProfiles in general.controller.js and getAvailableOotbActiveProfiles in groupingsService to call api in ui and store the available data in .availableProfiles

- Add ootb.active.user.profiles.json that contains multiple profiles information

- Apply ng-repeat with the values from scope.availableProfiles that contains the profiles from api call

- Add isOotb() function to check current spring boot active profile and apply to the menubar.html

- Add tests for loadAvailableProfiles and getAvailableOotbActiveProfiles in general.controller.test.js and groupings.service.test.js

- Enhance the Builder class functions in the User class (access folder)

- Refactor the file inputstream function in OotbActiveUserProfileService

- Implement getAvailableProfiles function test in the controller test, add @Mockbean OotbActiveUserProfileService and mock the map that contains the ootb active profiles

- Complete adding multiple ootb active profiles into the users map dynamically from json file with null safe and wrong path of file error
- Fix codacy code style
- Fix general controller test to check gs.getAvailableOotbActiveProfiles
- Modify the functions to initiate default active user by mapping JSON data to OotbActiveProfile with JsonUtil.asList() function
- Modify the end-point to get available profiles from OotbActiveUserProfileService
- Redesign the specification of JSON data for mapping to OotbActiveProfile
- Change the strategy to set property of Json file and the way to inject the file in the service layer
- Modify the tests for controller and service with dynamic active profile from JSON data
- Fix raw type of exception
- Implement the tests for ootb multiple active profiles in UserTest, RealmTest, JsonUtilTest, and OotbActiveProfileTest
# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/groupings/service/OotbActiveUserProfileService.java
+++ b/src/main/java/edu/hawaii/its/groupings/service/OotbActiveUserProfileService.java
@@ -1,51 +1,65 @@
 package edu.hawaii.its.groupings.service;
 
-import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
-import org.springframework.security.core.authority.AuthorityUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.stereotype.Service;
 
+import edu.hawaii.its.groupings.access.Role;
 import edu.hawaii.its.groupings.access.User;
+import edu.hawaii.its.groupings.type.OotbActiveProfile;
+import edu.hawaii.its.groupings.util.JsonUtil;
 
 @Service
 public class OotbActiveUserProfileService implements UserDetailsService {
-    private final Map<String, User> users = new HashMap<>();
+    private static final Log logger = LogFactory.getLog(OotbActiveUserProfileService.class);
+    private final Map<String, User> users = new LinkedHashMap<>();
+
+    // Default JSON file for active profiles
+    private String profilesFileName = "ootb.active.user.profiles.json";
 
     public OotbActiveUserProfileService() {
         initUsers();
     }
 
+    public OotbActiveUserProfileService(String profilesFileName) {
+        this.profilesFileName = profilesFileName;
+        initUsers();
+    }
+
     private void initUsers() {
-        // Initialize member with uid "member0123"
-        users.put("MEMBER", new User.Builder("member0123")  // UID is clearly the first parameter in the constructor
-                .uhUuid("11111111")
-                .authorities(AuthorityUtils.createAuthorityList("ROLE_UH", "ROLE_OOTB"))
-                .addAttribute("cn", "MEMBER")
-                .addAttribute("mail", "member@hawaii.edu")
-                .addAttribute("givenName", "DefaultMember")
-                .build());
+        List<OotbActiveProfile> ootbActiveProfiles =
+                JsonUtil.asList(JsonUtil.readJsonFileToString(profilesFileName), OotbActiveProfile.class);
+        ootbActiveProfiles.forEach(profile -> {
+            users.put(profile.getAttributes().get("givenName"), createUserFromProfile(profile));
+        });
+    }
 
-        // Initialize owner with uid "owner0123"
-        users.put("OWNER", new User.Builder("owner0123")
-                .uhUuid("22222222")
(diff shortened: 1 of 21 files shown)


mhodgesatuh (approved):
Looks okay AFAIK

src/main/resources/templates/menubar.html
```diff
@@ -59,15 +59,14 @@
                         </button>
                     </form>
                 </li>
-                <li class="nav-item dropdown" sec:authorize="isAuthenticated() and hasAuthority('ROLE_OOTB')">
+                <li class="nav-item dropdown" sec:authorize="isAuthenticated()" th:if="${@realm.isOotb()}" ng-click="loadAvailableProfiles()">
```
JorWo, line 62: You should be able to load the `ootb.active.user.profiles.json` file directly in Angular.js without having to go through an API endpoint. This should simplify the code a lot.

src/main/java/edu/hawaii/its/groupings/service/OotbActiveUserProfileService.java
```diff
@@ -1,51 +1,64 @@
...
+        List<String> roles = (List<String>) userData.get("authorities");
+        Map<String, String> attributes = (Map<String, String>) userData.get("attributes");
+
+        User.Builder builder = new User.Builder(uid)
+                .uhUuid(uhUuid)
+                .addAuthorities(roles);
+
+        attributes.forEach(builder::addAttribute);
+
+        return builder.build();
```
JorWo, line 61: I will say I do prefer your old code before the refactor because it was a lot simpler than the code that reads JSON files. You can make the call whether you think this refactor is needed or not.

I think your goal with this ticket was to put the default users inside a properties or JSON file. I would have gone with how you load data from the `data.harness` on the API side and put the JSON objects inside the application-ootb.properties or add a data.harness file to the UI side. Then use the JsonUtil class to load it. That way we can reuse the code for loading JSON files instead of adding new code here.

fduckart: Looks nice so far... but maybe using the Spring "@Resource" annotation would be good here. e.g., @Resource("${ootb.active.user.profiles.json}").  

The casts after reading the file are bit iffy. I think we have a helper method in the JsonUtil class to create objects. I think we should use that to create the User objects. If the helper method doesn't do exactly what you need, maybe you can create another helper method in JsonUtil.

gitCarrot: I also actually prefer the code before the refactor. However, during the last meeting, it was mentioned that the ootb project should serve as a repository with a variety of purposes beyond just showing the groupings project without the grouper client connection. The goal is for developers to be able to create various scenarios, use and test them directly in the development of groupings. During that discussion, it was mentioned that using JSON would make it easier for developers to view and modify the data, and easily import it with an overrides file. Additionally, when we later create something like a QA tool, it would be synergistic to have it in JSON format. Initially, I considered loading the JSON file into the properties file like the existing data harness method and calling it with propertyLocator, but I thought it would be difficult for developers to modify the data later if we compact the JSON data into a single line. So instead of properties, I decided to create it as a separate JSON file.

gitCarrot: > I will say I do prefer your old code before the refactor because it was a lot simpler than the code that reads JSON files. You can make the call whether you think this refactor is needed or not.
> 
> I think your goal with this ticket was to put the default users inside a properties or JSON file. I would have gone with how you load data from the `data.harness` on the API side and put the JSON objects inside the `application-ootb.properties` or add a `data.harness` file to the UI side. Then use the JsonUtil class to load it. That way we can reuse the code for loading JSON files instead of adding new code here.

I also actually prefer the code before the refactor. However, during the last meeting, it was mentioned that the ootb project should serve as a repository with a variety of purposes beyond just showing the groupings project without the grouper client connection. The goal is for developers to be able to create various scenarios, use and test them directly in the development of groupings. During that discussion, it was mentioned that using JSON would make it easier for developers to view and modify the data, and easily import it with an overrides file. Additionally, when we later create something like a QA tool, it would be synergistic to have it in JSON format. Initially, I considered loading the JSON file into the properties file like the existing data harness method and calling it with propertyLocator, but I thought it would be difficult for developers to modify the data later if we compact the JSON data into a single line. So instead of properties, I decided to create it as a separate JSON file.

JorWo:
Make sure to update the tests for User.java, JsonUtil.java, Realm.java, and add a set of tests for OotbActivePorfile.java, thanks!

JorWo (approved):

PR #953, opened 2024-07-09
Transition from Field Injection to Constructor Injection

hokwaichan:
# Ticket Link

[1719](https://uhawaii.atlassian.net/browse/GROUPINGS-1719)

# List of squashed commits

- Transition from Field Injection to Constructor Injection

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
@@ -8,7 +8,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.owasp.html.PolicyFactory;
 import org.owasp.html.Sanitizers;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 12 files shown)


src/main/java/edu/hawaii/its/groupings/access/AuthorizationService.java
```diff
@@ -17,13 +16,16 @@
...
 public class AuthorizationService {
 
-    @Autowired
-    GroupingsRestController groupingsRestController;
+    private final GroupingsRestController groupingsRestController;
 
-    @Autowired
-    UhUuidCheckerService uhUuidCheckerService;
+    private final UhUuidCheckerService uhUuidCheckerService;
     private static final Log logger = LogFactory.getLog(AuthorizationService.class);
```
fduckart, line 22: Can you move the logger to the top like we do in the other classes. Also, I'm not sure if we want the blank lines between the class-level variables, but whatever we decide is okay with me.

src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
```diff
@@ -30,26 +29,30 @@
...
-    private HttpRequestService httpRequestService;
+    private final HttpRequestService httpRequestService;
 
     // Constructor.
-    public OotbRestController() {
+    public OotbRestController(UserContextService userContextService,
+            OotbStaticUserAuthenticationFilter ootbAuthenticationFilter,
+            OotbActiveUserProfileService ootbActiveUserProfileService,
+            HttpRequestService httpRequestService) {
         policy = Sanitizers.FORMATTING;
```
fduckart, line 51: Unrelated to your PR, but I wonder if we can create this policy variable when we declare it above?

hokwaichan: I think the policy can be declared above

hokwaichan: However, I see most of the policies are declare inside the constructor, please let me know if we want to change it

JorWo: I think we should declare the `policy` above so that the constructor's main purpose is to inject services.

hokwaichan: I will move the policy above

JorWo (approved):
Looks good, just waiting on your API PR now.

PR #998, opened 2024-08-17
Added flag to prevent multiple triggers of getGroupingOwnersOnClick

jarellb:
# Ticket Link

[GROUPINGS-1765](https://uhawaii.atlassian.net/jira/software/c/projects/GROUPINGS/boards/18?assignee=712020%3Ad0e5c67f-f1f1-44a0-b5c9-282af528541e&selectedIssue=GROUPINGS-1765)

# List of squashed commits

1. Fixed codacy issue and shifted $scope.loadingOwner variables
2. Corrected coding standard issue

# Test Checklist

- [x] Exhibits Clear Code Structure:
- [x] Project Unit Tests Passed:
- [x] Project Jasmine Tests Passed:
- [x] Executes Expected Functionality:
- [x] Tested For Incorrect/Unexpected Inputs:

PR #1003, opened 2024-08-24
Add flags to prevent asynchronous triggers of addMembers

jarellb:
# Ticket Link

[Groupings-1770](https://uhawaii.atlassian.net/jira/software/c/projects/GROUPINGS/boards/18?assignee=712020%3Ad0e5c67f-f1f1-44a0-b5c9-282af528541e&selectedIssue=GROUPINGS-1770)

# List of squashed commits
* Added Jasmine tests for isAddingMembers guard
* Simplified Jasmine tests
* Moved spyOns to before scope.addMembers call

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -83,6 +83,7 @@
         $scope.membersInList = "";
         $scope.isMultiAdd = false;
         $scope.hasDeptAccount = false;
+        $scope.isAddingMembers = false;
 
         // Remove members
         $scope.multiRemoveResults = [];
@@ -666,12 +667,19 @@
          * @param {Object[]|null} uhIdentifiers
          */
         $scope.addMembers = (listName, uhIdentifiers) => {
+            // Prevent multiple asynchronous calls of addMembers
+            if ($scope.isAddingMembers) {
+                return;
+            }
+            $scope.isAddingMembers = true;
+
             // If uhIdentifiers parameter is null, get member input from $scope.manageMembers
             uhIdentifiers = $scope.sanitizer(uhIdentifiers ?? $scope.parseAddRemoveInputStr($scope.manageMembers));
             $scope.listName = listName;
             // Check if uhIdentifiers/member input is empty
             if (_.isEmpty(uhIdentifiers)) {
                 $scope.emptyInput = true;
+                $scope.isAddingMembers = false;
                 return;
             }
             // Prevent adding more than Threshold.MAX_IMPORT
@@ -679,6 +687,7 @@
                 $scope.displayDynamicModal(
                     Message.Title.IMPORT_OUT_OF_BOUNDS,
                     Message.Body.IMPORT_OUT_OF_BOUNDS);
+                $scope.isAddingMembers = false;
                 return;
             }
             // Check for members already in list, display error when all members to add already exist in the list
@@ -691,6 +700,7 @@
                 } else {
                     $scope.errorDismissed = false;
                 }
+                $scope.isAddingMembers = false;
                 return;
             }
 
@@ -706,6 +716,7 @@
 
             if (_.isEmpty(uhIdentifiers)) {
                 $scope.containsInput = true;
+                $scope.isAddingMembers = false;
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


JorWo:
I see duplicate code changes from a previous PR. Running `git pull upstream --rebase main` then force pushing the changes back up should do the trick.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -666,19 +667,27 @@
          * @param {Object[]|null} uhIdentifiers
          */
         $scope.addMembers = (listName, uhIdentifiers) => {
+            // Prevent multiple asynchronous calls of addMembers
+            if ($scope.addingMembers === true) {
+                return;
+            }
```
JorWo, line 673: When naming booleans, we tend to prefix it with "is" so that it is more clean to read when doing the following:
```
if ($scope.isAddingMembers) {
    return;
}
```
That way, you do not need to specifically check for `=== true`. This can be more dangerous in Javascript without any typing restrictions like Typescript has, but we can definitely trust a variable to be a boolean with that naming convention.

src/test/javascript/grouping.controller.test.js
JorWo, line 1: A new test should be created to specifically check that certain functions are not being called and certain variables are not being updated when $scope.addingMembers is true.

src/main/resources/static/javascript/mainApp/grouping.controller.js
jarellb: Should I also do this for my previous ticket of the same issue? https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/pull/998

JorWo: Yeah good catch, you can include the change in this ticket. I guess you could keep the variable name $scope.loadingOwners the same because we already have $scope.loading everywhere instead of $scope.isLoading.

src/test/javascript/grouping.controller.test.js
```diff
@@ -918,6 +920,42 @@ describe("GroupingController", () => {
...
+            spyOn(scope, 'displayApiErrorModal');
+
+            expect(scope.displayDynamicModal).not.toHaveBeenCalled();
+            expect(scope.existsInList).not.toHaveBeenCalled();
+            expect(gs.getMemberAttributeResultsAsync).not.toHaveBeenCalled();
+            expect(gs.getMemberAttributeResults).not.toHaveBeenCalled();
+            expect(scope.displayImportErrorModal).not.toHaveBeenCalled();
+            expect(scope.displayImportConfirmationModal).not.toHaveBeenCalled();
+            expect(scope.displayAddModal).not.toHaveBeenCalled();
+            expect(scope.displayApiErrorModal).not.toHaveBeenCalled();
```
JorWo, line 956: Sorry I should have been more specific, I would only check that most important function was not called. In this case, it would be `gs.getMemberAttributeResultsAsync` and `gs.getMemberAttributeResults` becuase it invokes an API call. A problem we have with our unit tests is that they are extremely fragile where editing one line of code just breaks a bunch of tests. Just keeping it minimal here is the best approach.

src/test/javascript/grouping.controller.test.js
```diff
@@ -918,6 +920,42 @@ describe("GroupingController", () => {
...
+            scope.addMembers("Include", uhIdentifiers);
+
+            expect(scope.listName).toBeUndefined();
+            expect(scope.errorDismissed).toBeUndefined();
+            expect(scope.isBatchImport).toBeUndefined();
+            expect(scope.containsInput).toBeUndefined();
+            expect(scope.invalidMembers).toEqual([]);
+            expect(scope.addInputError).toBeUndefined();
+            expect(scope.waitingForImportResponse).toBeFalse();
+        });
```
JorWo, line 934: Following the previous comment, this test can be removed, thanks.

src/test/javascript/grouping.controller.test.js
```diff
@@ -918,6 +920,17 @@ describe("GroupingController", () => {
             expect(scope.isBatchImport).toBeTrue();
         });
 
+        it("should not call functions within $scope.addMembers if $scope.isAddingMembers is true", () => {
+            scope.isAddingMembers = true;
+            scope.addMembers("Include", uhIdentifiers);
+
+            spyOn(gs, 'getMemberAttributeResultsAsync');
+            spyOn(gs, 'getMemberAttributeResults');
```
JorWo, line 928: This completely went over my head but the spyOn's should be above `scope.addMembers()` because you want to start spying before you call the function. Otherwise the test will always pass if `scope.isAddMembers` was false.

JorWo (approved):
Looks great.

PR #1016, opened 2024-09-05
Improve ootb user experience and fix minor errors

gitCarrot:
# Ticket Link

[Ticket-1773](https://uhawaii.atlassian.net/browse/GROUPINGS-1773)

# List of squashed commits

- Add post construct on the OotbRestController to post update active profile with admin profiles
- Add the logic to redirect to the page to reload the page when user at home in ootb to avoid cache issue when changing active profiles
- Enhance the data for active user profiles in json file
- Modify all the ootb test class with mocking ootb controller to return null when it interacts with api projects in the class
- Change coverage yml with upload-artifact@v4
- Fix MockMVC to test end point for ootb rest controller
- Change dependency injections between ootb related classes by removing some beans and adding some annotations and local field in the class
- Modify the test for the changes in the class and add @ActiveProfiles annotation to separate the test from localhost env
- Fix codacy code style- 
- Add and Test various scenario with adding basis grouping member and add some logic to make difference when adding member between include and basis extension
- Add missed logic to set currentUser logged in into the userDetail in the ootbAuthenticationFilter when updateActiveDefaultUser end-point is executed

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
@@ -16,7 +16,6 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
-import edu.hawaii.its.api.service.HttpRequestService;
 import edu.hawaii.its.api.service.OotbHttpRequestService;
 import edu.hawaii.its.groupings.access.UserContextService;
 import edu.hawaii.its.groupings.configuration.OotbStaticUserAuthenticationFilter;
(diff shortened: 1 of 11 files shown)


mhodgesatuh (commented):
Over my head.  I'll leave this entirely to Jordan.

.github/workflows/coverage.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           git push
         fi
     - name: Upload Jacoco coverage report
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
```
JorWo, line 55: Could you run `git pull upstream --rebase main` so that this change from a previous PR does not appear in this PR? Thanks.

src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
```diff
@@ -15,40 +18,50 @@
...
+    @PostConstruct()
+    public void init() {
+        if(realmService.isOotb()) {
+            String adminProfile = ootbActiveUserProfileService.findGivenNameForAdminRole();
+            if (adminProfile != null) {
+                updateActiveDefaultUser(adminProfile);
+            } else {
+                logger.error("Admin profile not found, unable to initiate default API request.");
+            }
+        }
```
JorWo, line 64: This seems to be something that should be called inside a configuration class rather than in a controller.

.github/workflows/coverage.yml
gitCarrot: I'll keep in mind before pushing origin! Thanks

src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
gitCarrot: I reorganize the configuration and dependency injections for now to make it call inside of a configuration class!

gitCarrot:
Sorry I need to fix several issues. It is going to be done soon

JorWo:
> Sorry I need to fix several issues. It is going to be done soon

Sounds good!

JorWo (approved):
Looks good to me!

PR #1018, opened 2024-09-13
Fix authentication filter conflict in security config in localhost environment

gitCarrot:
# Ticket Link

[Ticket-1784](https://uhawaii.atlassian.net/browse/GROUPINGS-1784)

# List of squashed commits

- Fix authentication filter conflict in security config in localhost environment

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
@@ -7,6 +7,7 @@ import org.apache.commons.logging.LogFactory;
 import org.owasp.html.PolicyFactory;
 import org.owasp.html.Sanitizers;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Profile;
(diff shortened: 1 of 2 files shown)


mhodgesatuh (approved):
looks extremely straightforward.   We have unit testing to ensure it works?

gitCarrot:
To be honest, there should have been a test for this bean in securityConfig to detect this problem , but now I deleted the bean to test, so I don't need a unit test.

The reason there was no securityConfig test was probably because it was difficult to test an automated securityConfig and not easy to mock security object. Maybe that's why Frank only implemented construction test too.

gitCarrot:
The reason why I made OotbStaticUserAuthenticationFilter bean in both config class was that ootb controller that used it was registered as bean by component scan in spring boot when user is on localhost environment since there was no profile annotation "ootb" for ootb controller. So I thought I had to initialize the bean in secureConfig class too at least, but it was not if I use @profile annotation in controller.

JorWo (approved):

PR #1045, opened 2024-11-05
Investigate and fix the possible issue from spring boot devtools maven dependency

gitCarrot:
# Ticket Link

[Ticket-1850](https://uhawaii.atlassian.net/browse/GROUPINGS-1850)

# List of squashed commits

- Investigate and fix the possible issue from spring boot devtools maven dependency
- Make configuration class for hot reload on localhost without setting absolute project directory path in application.properties
- Implement the test checking the config class for hot reload
- Fix codacy code style

# Test Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,13 +77,6 @@
             <artifactId>spring-boot-starter-actuator</artifactId>
         </dependency>
 
-        <!-- hot swapping, disable cache for template, enable live reload -->
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-devtools</artifactId>
-            <optional>true</optional>
-        </dependency>
-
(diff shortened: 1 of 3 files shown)


mhodgesatuh (approved):

JorWo:
We should keep this dependency because live reload is nice to have for Angular.js.

gitCarrot:
Are we actually using live reload now in angular js? Without adding some devtools properties, it seems live reload is not working on current configuration

JorWo:
Yeah we have used it in the past but it seems that is has broke after updating the Spring Boot version. I remember the live reload with Angular.js only works with .js files and not .html. Could you investigate further? Thanks.

mhodgesatuh (approved):

JorWo (approved):

PR #1056, opened 2025-02-09
Log the parameters in GroupingsRestController

rpirnia:
# Ticket Link

GROUPINGS 1903: https://uhawaii.atlassian.net/browse/GROUPINGS-1903

# List of squashed commits

- Commit 1

# Test Checklist

- [X] Project Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
@@ -122,6 +122,9 @@ public class GroupingsRestController {
             @RequestParam(required = true) String sortBy,
             @RequestParam(required = true) Boolean isAscending) {
         logger.info("Entered REST getGrouping...");
+        String groupPathsStr = groupPaths != null ? groupPaths.toString() : "null";
+        logger.info(String.format("Received Parameters - groupPaths: %s, page: %d, size: %d, sortBy: %s, isAscending: %b",
+                groupPathsStr, page, size, sortBy, isAscending));
         String currentUid = policy.sanitize(userContextService.getCurrentUid());
         Map<String, String> params = mapGroupingParameters(page, size, sortBy, isAscending);
         String baseUri = API_2_1_BASE + "/groupings/group";
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
@@ -100,16 +100,16 @@ public class GroupingsRestController {
 
     @GetMapping(value = "/groupings/admins")
     public ResponseEntity<String> groupingAdmins() {
-        logger.info("Entered REST groupingAdmins...");
         String currentUid = policy.sanitize(userContextService.getCurrentUid());
+        logger.info(String.format("Entered REST groupingAdmins - currentUid: %s", currentUid));
         String uri = API_2_1_BASE + "/groupings/admins";
         return httpRequestService.makeApiRequest(currentUid, uri, HttpMethod.GET);
     }
 
     @GetMapping(value = "/groupings")
     public ResponseEntity<String> allGroupings() {
-        logger.info("Entered REST allGroupings...");
         String currentUid = policy.sanitize(userContextService.getCurrentUid());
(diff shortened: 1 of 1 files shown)


mhodgesatuh (commented):
Is this mod going to log each page viewed by the owner?  If so, that seems excessive given the original vision, which was merely to show which grouping an owner was currently working with when they initially switched to that grouping.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
```diff
@@ -122,6 +122,9 @@ public ResponseEntity<String> getGrouping(@RequestBody(required = true) List<Str
             @RequestParam(required = true) String sortBy,
             @RequestParam(required = true) Boolean isAscending) {
         logger.info("Entered REST getGrouping...");
+        String groupPathsStr = groupPaths != null ? groupPaths.toString() : "null";
+        logger.info(String.format("Received Parameters - groupPaths: %s, page: %d, size: %d, sortBy: %s, isAscending: %b",
+                groupPathsStr, page, size, sortBy, isAscending));
         String currentUid = policy.sanitize(userContextService.getCurrentUid());
```
JorWo, line 128: Could you include the currentUid in the logger message, preferably at the beginning? 

Also it would be beneficial to log the parameters of the rest of the endpoints.

Thanks!

src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
```diff
@@ -122,6 +122,9 @@ public ResponseEntity<String> getGrouping(@RequestBody(required = true) List<Str
             @RequestParam(required = true) String sortBy,
             @RequestParam(required = true) Boolean isAscending) {
         logger.info("Entered REST getGrouping...");
+        String groupPathsStr = groupPaths != null ? groupPaths.toString() : "null";
```
JorWo, line 125: Correct me if I am wrong, you don't need to check for null or convert groupPaths to a string since the logger and String.format functions will handle those cases.

JorWo:
> Is this mod going to log each page viewed by the owner? If so, that seems excessive given the original vision, which was merely to show which grouping an owner was currently working with when they initially switched to that grouping.

That is a bit tough to accomplish because you cannot log from Angular.js into the console. Maybe putting the parameters received in the same logger of "Entered REST getGrouping..." would keep the logs compact.

mhodgesatuh:
> > Is this mod going to log each page viewed by the owner? If so, that seems excessive given the original vision, which was merely to show which grouping an owner was currently working with when they initially switched to that grouping.
> 
> That is a bit tough to accomplish because you cannot log from Angular.js into the console. Maybe putting the parameters received in the same logger of "Entered REST getGrouping..." would keep the logs compact.

+1 for putting the additional info on the same line.

rpirnia:
Great, I'll make those changes today.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestController.java
```diff
@@ -131,24 +132,26 @@ public ResponseEntity<String> getGrouping(@RequestBody(required = true) List<Str
...
+        logger.info(String.format("Entered REST getGroupingDescription - currentUid: %s, groupPath: %s",
+                currentUid, groupPath));
         String uri = String.format(API_2_1_BASE + "/groupings/%s/description", policy.sanitize(groupPath));
         return httpRequestService.makeApiRequest(currentUid, uri, HttpMethod.GET);
     }
 
     @GetMapping(value = "/groupings/{groupPath}/groupings-sync-destinations")
     public ResponseEntity<String> getGroupingSyncDest(Principal principal, @PathVariable String groupPath) {
-        logger.info("Entered REST getGroupingSyncDest...");
+        logger.info(String.format("Entered REST getGroupingSyncDest - groupPath: %s", groupPath));
```
JorWo, line 144: Missed principalName here.

rpirnia: Thank you for catching that.

JorWo (approved):
Looks good!

PR #1059, opened 2025-02-26
Add owner grouping logic to add/remove owners

lukepag:
# Ticket Link

[1682](https://uhawaii.atlassian.net/browse/GROUPINGS-1682)

# List of squashed commits

- Add owner grouping logic to add/remove owners
- add new modals for adding and removing group path owners
- Create new controller functions
- Implement new grouping owner add/remove logic
- Fixed add bug
- Implement hyperlink for grouping owners
- Add more comments and fix CSS spacing
- Create and update existing jasmine tests
- Finalize clean up

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -85,6 +85,12 @@
         $scope.hasDeptAccount = false;
         $scope.isAddingMembers = false;
 
+        // Always initially set to default modal
+        $scope.addModalId = "add-modal";
+        $scope.addModalURL = "modal/addModal";
+        $scope.removeModalId = "remove-modal";
+        $scope.removeModalURL = "modal/removeModal";
+
         // Remove members
         $scope.multiRemoveResults = [];
         $scope.membersToRemove = [];
@@ -113,6 +119,29 @@
          */
         $scope.displayGrouping = (currentPage, index) => {
             $scope.selectedGrouping = $scope.pagedItemsGroupings[Number(currentPage)][Number(index)];
+            $scope.getGroupingInformation();
+            $scope.showGrouping = true;
+        };
+
+        /**
+         * Called when an owner-grouping hyperlink is clicked
+         * Saves the base path and grouping name of the owner-grouping in "selectedOwnerGrouping" and opens a new tab
+         * @param {String} basePath - the base path of the owner-grouping clicked by the user
+         * @param {String} groupingName - the name of the owner-grouping clicked by the user
+         */
+        $scope.displayOwnerGroupingInNewTab = (basePath, groupingName) => {
+            sessionStorage.setItem("selectedOwnerGrouping", JSON.stringify({ path: basePath, name: groupingName }));
+            $window.open("groupings");
+            sessionStorage.removeItem("selectedOwnerGrouping");
+        };
+
+        /**
+         * Retrieves the selectedOwnerGrouping data from displayOwnerGroupingInNewTab and displays the grouping in the new tab
+         * @param selectedOwnerGrouping - object that holds the path and name of the selected owner-grouping
+         */
+        $scope.displayOwnerGrouping = (selectedOwnerGrouping) => {
+            $scope.selectedGrouping = selectedOwnerGrouping;
+            $scope.currentPageOwners = 0;
             $scope.getGroupingInformation();
             $scope.showGrouping = true;
         };
@@ -166,7 +195,7 @@
         $scope.getGroupingInformation = async () => {
             loadMembersList = true;
             const groupingPath = $scope.selectedGrouping.path;
-            const paths = [groupingPath + ":basis", groupingPath + ":include", groupingPath + ":exclude", groupingPath + ":owners"];
+            const paths = [groupingPath + ":basis", groupingPath + ":include", groupingPath + ":exclude"];
             let currentPage = 1;
             $scope.setPage('First', 'currentPageMembers', 'pagedItemsMembers');
             $scope.loading = true;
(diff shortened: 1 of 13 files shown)


mhodgesatuh (commented):
* Give ownership of grouping at grouping path to newOwner. A user with owner privileges has
     * read and write privileges
     * of a grouping.
     
     The above should be on two lines.
     
     In the comments the spelling "owners-group" might be clearer.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -234,6 +266,43 @@
...
+                    $scope.groupingOwners.forEach((owner) => {
+                        // Normal member owners cannot have colons in their name
+                        if(owner.name.includes(':')) {
+                            owner.isOwnerGrouping = true;
+                            // Name field in owner groupings initially holds the group path
+                            owner.ownerGroupingPath = owner.name;
+                            const splitOwnerPath = owner.name.split(':');
+                            // Isolate the grouping name from the path
+                            owner.name = splitOwnerPath[splitOwnerPath.length - 1];
+                        }
```
JorWo, line 289: This logic should be on the API side, but that can be for another ticket.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -637,23 +707,53 @@
...
         $scope.addOnClick = (listName) => {
             $scope.resetErrors();
-            if (listName === "Include" || listName === "Exclude" || listName === "owners") {
+            if (listName === "Include" || listName === "Exclude") {
+                $scope.addMembers(listName);
+            }
+            if (listName === "owners") {
+                // If the user input has a colon, we can assume that it's a group path
+                if ($scope.manageMembers.includes(':')) {
+                    listName = "owner grouping";
```
JorWo, line 720: There technically is no listName/groupName for owner groupings, it would still be the owner group. Could you use a $scope.isOwnerGrouping variable instead?

JorWo:
Are the cases handled when adding:
1. More than one owner grouping path
2. A mix of uhIdentifiers and owner grouping paths in the input box?

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -624,7 +694,7 @@
         $scope.existsInList = (listName, members) => {
             const currentPage = getCurrentPage(listName);
             const membersInList = members.filter((member) =>
-                _.some(currentPage, {uhUuid: member}) || _.some(currentPage, {uid: member})
+                _.some(currentPage, {uhUuid: member}) || _.some(currentPage, {uid: member}) || _.some(currentPage, {name: $scope.groupingName})
```
JorWo, line 697: Just a question, what does the `|| _.some(currentPage, {name: $scope.groupingName}` do?

src/main/resources/templates/modal/addOwnerGroupingModal.html
```diff
@@ -0,0 +1,47 @@
...
+        <span class="font-weight-bold">{{groupingName}}</span>
+        <span class="modal-black-text">to the owners list?</span>
+    </p>
+    <!--
+    <p class="alert alert-warning mt-2" role="alert" ng-if="hasDeptAccount">
+        <span class="text-dark" th:text="#{screen.message.modal.add.deptAccount}"></span>
+    </p>
+
+    do I need to account for departmental accounts that are in the grouping?
+    -->
```
JorWo, line 37: I noticed you left this comment here.

JorWo:
Make sure to fix the Codacy issues, thanks.

src/main/resources/static/javascript/mainApp/grouping.controller.js
lukepag: @JorWo One slight issue I see with this is that the owner objects have the owner.isOwnerGrouping field and that may be confusing to some. However, if this isn't an issue I can push up the changes!

lukepag:
> Are the cases handled when adding:
> 
> 1. More than one owner grouping path?
> 2. A mix of uhIdentifiers and owner grouping paths in the input box?

I believe I had a talk with Michael about how we don't need to account for situations where users are adding more than one grouping path. @mhodgesatuh could I get your opinion on this?

Regarding the second question, it does not currently handle the mix of uhIdentifiers and owner grouping paths

src/main/resources/static/javascript/mainApp/grouping.controller.js
JorWo: There are many places that in the code that use a $scope and non-scope variable to describe the same thing like $scope.listName and listName, so I assume both owner.isOwnerGrouping and $scope.isOwnerGrouping being used would not be too confusing. 

Having checks for `listName == "owners" && isOwnerGrouping` is fine because it is still readable.

mhodgesatuh:
Regarding the more than one grouping path question, here's some context: 
The feature will only be used by select CSOC individuals and will not be advertised to our broader audience. So the thought is that for this first iteration of the feature it is okay if we don't have rigorous user input testing. If we encounter issues, a future iteration could introduce more controls and make the feature higher profile.

JorWo:
> Regarding the more than one grouping path question, here's some context: The feature will only be used by select CSOC individuals and will not be advertised to our broader audience. So the thought is that for this first iteration of the feature it is okay if we don't have rigorous user input testing. If we encounter issues, a future iteration could introduce more controls and make the feature higher profile.

@mhodgesatuh Any thoughts on displaying a "feature not supported" modal if those select CSOC individuals did attempt to more than one owner that could be a mix of uhIdentifiers and groupingPaths? I want to make sure there are no disastrous consequences of not handling those cases.

mhodgesatuh:
@JorWo Please update the ticket to describe at least include a minimal set of input controls to ensure that there is no unanticipated behavior. I'm fine, for example, if the initial implementation only allows for a single owner-grouping reference, in order to get minimal functionality established.

JorWo:
> @JorWo Please update the ticket to describe at least include a minimal set of input controls to ensure that there is no unanticipated behavior. I'm fine, for example, if the initial implementation only allows for a single owner-grouping reference, in order to get minimal functionality established.

Ok, I added a user input edge cases section to the bottom of the ticket.

src/main/resources/static/javascript/mainApp/grouping.controller.js
lukepag: @JorWo Got it! I'll make the changes

mhodgesatuh (commented):
I suggest we consistent hyphenate owner-grouper to represent the use of a grouping to indicate the indirect owners, AKA co-owners.

lukepag:
> I suggest we consistent hyphenate owner-grouper to represent the use of a grouping to indicate the indirect owners, AKA co-owners.

Are you suggesting we do this for every instance of "owner grouping" that we currently have? This seems like a good way to set the distinction

mhodgesatuh:
> Are you suggesting we do this for every instance of "owner grouping" that we currently have? This seems like a good way to set the distinction

Indeed I am.  I believe it adds clarity where otherwise the reference can be confusing, especially for less-experienced team members.

JorWo (approved):
Looks good!

PR #1071, opened 2025-03-29
Prevent multi-add for owner-groupings

lukepag:
# Ticket Link

[1937](https://uhawaii.atlassian.net/browse/GROUPINGS-1937)

# List of squashed commits

- Prevent multi-add for owner-groupings

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -708,14 +708,15 @@
          * @param {String} listName grouping list (i.e. include, exclude, owners, or owner-grouping)
          */
         $scope.addOnClick = (listName) => {
+            const userInput = $scope.parseAddRemoveInputStr($scope.manageMembers);
             $scope.resetErrors();
             if (listName === "Include" || listName === "Exclude") {
+                // Prevents adding owner-groupings to include/exclude list
                 if ($scope.manageMembers.includes(":")) {
-                    $scope.displayDynamicModal(Message.Title.NO_MEMBERS_ADDED, Message.Body.ADD_GROUP_PATH_ERROR);
+                    $scope.displayDynamicModal(Message.Title.BAD_INPUT_ERROR, Message.Body.ADD_INPUT_ERROR);
                     clearMemberInput();
-                    return;
                 } else {
-                    $scope.addMembers(listName);
+                    $scope.addMembers(listName, userInput);
                 }
             }
             if (listName === "owners") {
@@ -726,7 +727,7 @@
                         $scope.displayDynamicModal(Message.Title.OWNER_NOT_ADDED, Message.Body.ADD_CURRENT_PATH_ERROR);
                         clearMemberInput();
                         return;
-                    } else if ($scope.manageMembers.includes(",")) {
+                    } else if (userInput.length > 1) {
                         // Prevent multi-adding owner-groupings
                         $scope.displayDynamicModal(Message.Title.INVALID_MULTI_ADD, Message.Body.INVALID_MULTI_ADD);
                         clearMemberInput();
@@ -743,7 +744,7 @@
                     $scope.addModalId = "add-modal";
                     $scope.addModalURL = "modal/addModal";
                 }
-                $scope.addMembers(listName);
+                $scope.addMembers(listName, userInput);
             }
             $scope.errorDismissed = false;
         };
@@ -753,6 +754,7 @@
          * @param {String} listName grouping list (i.e. Include, Exclude, owners, or owner-grouping)
          */
         $scope.removeOnClick = (listName) => {
+            const userInput = $scope.parseAddRemoveInputStr($scope.manageMembers);
             $scope.resetErrors();
             if (listName === "Include" || listName === "Exclude") {
                 $scope.removeMembers(listName);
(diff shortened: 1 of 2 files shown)


src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -726,7 +727,7 @@
                         $scope.displayDynamicModal(Message.Title.OWNER_NOT_ADDED, Message.Body.ADD_CURRENT_PATH_ERROR);
                         clearMemberInput();
                         return;
-                    } else if ($scope.manageMembers.includes(",")) {
+                    } else if ($scope.manageMembers.includes(", ")) {
```
JorWo, line 730: Does that work the same as `if ($scope.manageMembers.includes(",") || $scope.manageMembers.includes(","))`? The goal is to catch "a,b,c" or "a b c" when adding to a group.

lukepag: Ah I see, so the three valid multi add forms are "a, b, c" "a b c" and "a,b,c" ?

mhodgesatuh:
# Ah I see, so the three valid multi add forms are "a, b, c" "a b c" and "a,b,c" ?
The pattern I've seen over the years is to first sanitize the user input, and then process it.  In the example above the commas would be replaced with spaces and multiple spaces would be compressed to a single space.  And then the sanitized input would be processed with simple logic.  Note that since this is user input, sanitization also includes ensuring only valid characters are provided, etc, which should already be in the code, but worth double checking.

mikhail-shkaralevich:
> # Ah I see, so the three valid multi add forms are "a, b, c" "a b c" and "a,b,c" ?
> The pattern I've seen over the years is to first sanitize the user input, and then process it. In the example above the commas would be replaced with spaces and multiple spaces would be compressed to a single space. And then the sanitized input would be processed with simple logic. Note that since this is user input, sanitization also includes ensuring only valid characters are provided, etc, which should already be in the code, but worth double checking.

Yes, I agree. It does sound like a good idea to modify user's input to the standardized form. We do sanitize user's input in the UI controller before sending data to the API, so i am not sure if it is worth doing this in the Angular project since we are switching to React in the near future. The react project should definitely have this process in place.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -708,8 +708,10 @@
          * @param {String} listName grouping list (i.e. include, exclude, owners, or owner-grouping)
          */
         $scope.addOnClick = (listName) => {
+            let userInput = $scope.parseAddRemoveInputStr($scope.manageMembers);
```
JorWo, line 711: This should be const if it's not being redefined else-where, thanks.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -753,19 +755,27 @@
          * @param {String} listName grouping list (i.e. Include, Exclude, owners, or owner-grouping)
          */
         $scope.removeOnClick = (listName) => {
+            let userInput = $scope.parseAddRemoveInputStr($scope.manageMembers);
```
JorWo, line 758: Same for here.

JorWo:
Functionality-wise, with the sanitizer at the start, this looks good!

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -753,19 +755,27 @@
...
-                    $scope.removeModalId = "remove-owner-grouping-modal";
-                    $scope.removeModalURL = "modal/removeOwnerGroupingModal";
-                    // ownerGroupPath is specifically used for the path in the removeOwnerGroupingModal
-                    $scope.ownerGroupPath = $scope.manageMembers;
-                    $scope.groupingName = $scope.manageMembers.split(":").pop();
+                    if (userInput.length > 1) {
+                        // Prevent multi-adding owner-groupings
+                        $scope.displayDynamicModal(Message.Title.INVALID_MULTI_REMOVE, Message.Body.INVALID_MULTI_REMOVE);
+                        clearMemberInput();
+                        return;
```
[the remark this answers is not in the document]
lukepag: Got it!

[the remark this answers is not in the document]
mikhail-shkaralevich: Line 733*

mikhail-shkaralevich:
@lukepag, how's the progress with this PR? It needs a small adjustment and it is ready to be merged! Everything else looks good.

lukepag:
> @lukepag, how's the progress with this PR? It needs a small adjustment and it is ready to be merged! Everything else looks good.

I'm sorry, but I believe I fixed all the requested changes. Am I missing something?

mikhail-shkaralevich:
> > @lukepag, how's the progress with this PR? It needs a small adjustment and it is ready to be merged! Everything else looks good.
> 
> I'm sorry, but I believe I fixed all the requested changes. Am I missing something? I forced pushed up all the changes last month which you can see in the PR history. Lmk if you can see it.

Yes, the last change needs to be made is to remove the unnecessary return statements. Please see Jordan's and mine comment above. Thanks!

lukepag:
> > > @lukepag, how's the progress with this PR? It needs a small adjustment and it is ready to be merged! Everything else looks good.
> > 
> > 
> > I'm sorry, but I believe I fixed all the requested changes. Am I missing something? I forced pushed up all the changes last month which you can see in the PR history. Lmk if you can see it.
> 
> Yes, the last change needs to be made is to remove the unnecessary return statements. Please see Jordan's and my comment above. Thanks!

I believe we need these return statements. If they were not included, it would continue to call the scope function to add/remove the specified member, since the function call is outside of the if statements

mikhail-shkaralevich (approved):

PR #1086, opened 2025-06-10
Dev 1953 rainemendoza

rainemendoza:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/jira/software/c/projects/GROUPINGS/boards/18?quickFilter=75&selectedIssue=GROUPINGS-1953)

# List of squashed commits

- Fixed the bug when removing yourself. Correct Modal now appears
- add the text "you will be redirected back to home page" to the modal to let user know after they confirm the romoval

# Test Checklist

### DIFF
--- a/src/main/resources/static/javascript/mainApp/grouping.controller.js
+++ b/src/main/resources/static/javascript/mainApp/grouping.controller.js
@@ -1466,14 +1466,9 @@
          * returns false
          */
         $scope.showWarningRemovingSelf = () => {
-            if (!$scope.membersToRemove || !$scope.currentUser) {
-                return false;
-            }
-            const removingSelf = $scope.membersToRemove.includes($scope.currentUser.uhUuid);
-            const isSensitiveList = $scope.listName === "admins" || $scope.listName === "owners";
-
-
-            return removingSelf && isSensitiveList;
+            return ($scope.membersToRemove.includes($scope.currentUser.uid)
+                    || $scope.membersToRemove.includes($scope.currentUser.uhUuid))
+                    && ($scope.listName === "owners" || $scope.listName === "admins");
         };
 
         /**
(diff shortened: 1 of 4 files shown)


mikhail-shkaralevich:
@rainemendoza , did you run the test? If so, please check the corresponding checkboxes for the PR.

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -1466,9 +1466,14 @@
          * returns false
          */
         $scope.showWarningRemovingSelf = () => {
-            return ($scope.membersToRemove.includes($scope.currentUser.uid)
-                    || $scope.membersToRemove.includes($scope.currentUser.uhUuid))
-                && ($scope.listName === "owners" || $scope.listName === "admins");
+            if (!$scope.membersToRemove || !$scope.currentUser) {
+                return false;
+            }
+            const removingSelf = $scope.membersToRemove.includes($scope.currentUser.uhUuid);
```
mikhail-shkaralevich, line 1472: It can be either a UID or UHUID

src/main/resources/static/javascript/mainApp/grouping.controller.js
```diff
@@ -1466,9 +1466,14 @@
          * returns false
          */
         $scope.showWarningRemovingSelf = () => {
-            return ($scope.membersToRemove.includes($scope.currentUser.uid)
-                    || $scope.membersToRemove.includes($scope.currentUser.uhUuid))
-                && ($scope.listName === "owners" || $scope.listName === "admins");
+            if (!$scope.membersToRemove || !$scope.currentUser) {
```
mikhail-shkaralevich, line 1469: This is not a proper check. It shouldn't be the case that either membersToRemove or currentUser is not set when a user attempts to remove someone from a list. Is there a case where membersToRemove or currentUser is not set? If so, then that part of the code needs to fixed.

rainemendoza: I reverted it back to what it originally was because it seems the issue was more with there was no call for the current user so that was the big thing that was needed. The original code worked as intended but just needed that variable.

PR #1171, opened 2026-04-23
Upgrade Thymeleaf version

rpirnia:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-2173)

# List of squashed commits

- Upgrade Thymeleaf version

# Test Checklist

- [X] Project Unit Tests Passed:
- [X] Checked For ADA Compliance:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
         <angularjs-version>1.8.3</angularjs-version>
         <popper.js-version>1.16.1-lts</popper.js-version>
         <bootstrap-version>4.6.2</bootstrap-version>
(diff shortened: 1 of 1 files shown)


mhodgesatuh (commented):
This worries me because it appears that Thymeleaf is a new entry, which suggests that a parent package previously invoke it.  Is that the case and if so which is the parent.

mhodgesatuh (commented):
Looks like we are adding to the pom file rather than just updating it. Normally, we only update it. I might be missing something though.

